PR #39, opened 2014-05-05
convert parsing to boost::program_options

### LINKED ISSUE #37: Use boost program options
Use [boost program options](https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/src/libPMacc/examples/gameOfLife2D/main.cu) to define program parameter in the examples.

slizzered:
close #37

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ endif()
 ###############################################################################
 # Boost
 ###############################################################################
-find_package(Boost 1.48.0 REQUIRED)
+find_package(Boost 1.48.0 REQUIRED COMPONENTS program_options)
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${Boost_LIBRARIES})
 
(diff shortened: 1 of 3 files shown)


ax3l:
-> should be optional and only for tests (e.g. compile tests only when boost is available)

CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ endif()
 ###############################################################################
 # Boost
 ###############################################################################
-find_package(Boost 1.48.0 REQUIRED)
+find_package(Boost 1.48.0 REQUIRED COMPONENTS program_options)
```
ax3l: before your "verify heap test" sections, run
```cmake
find_package(Boost 1.48.0 COMPONENTS program_options)
```
*again* and branch on
```cmake
if(Boost_PROGRAM_OPTIONS_FOUND)
  set(LIBS ${LIBS} ${Boost_PROGRAM_OPTIONS_LIBRARY})
  # ... cuda_add_executable, etc.
endif()
```

ax3l:
having only header boost dependencies is actually a good thing compared to also compiled dependencies, so we close this for now.

Issue #72, opened 2014-11-25
mallocMC new


psychocoderHPC:
We need a own mallocMC::new().

If a allocator use the default new operator we can't overwrite them. This means we can't write once the code and test with different mallocMC allocator models.

slizzered:
Directly using `new` inside a namespace is not possible (C++ forbids it).
The code already allows to use `mallocMC::malloc()` with the classic C-style syntax.
My suggestion is to remove the offending declarations completely (no longer overwrite the default new/malloc function from CUDA). This will avoid conflicts and make mallocMC-code more explicit. And no new code has to be added.

Remaining problem: `malloc()` does not give you the type without an explicit cast. the C++ `new` operator uses some internal magic to make this work. Maybe there is a way to do the same with templates...

PR #75, opened 2015-01-15
Edited changelog: removed #72 from release 2.0.1

slizzered:
- #72 will not make it into the release and is pushed back to the
   release of 2.1.0crp.


ax3l:
:+1: thx!

PR #79, opened 2015-01-22
fixed wrong calculation of segments

slizzered:
## TL;DR
`fullsegments` and/or `additional_chunks` could get way too large in some situations.

## Full text:
Usually, the `pagesize` is chosen, so that at most 1024 chunks fit inside a single page. If there are more than 32 chunks, they are tracked with special bitmasks on the page itself (`onpagemasks`). Each `onpagemask` tracks one `segment`. Usually, a segment contains 32 chunks and is called a `fullsegment`. The last segment sometimes holds less than 32 chunks, since there is not enough space left. These few chunks are called `additional_chunks`. There can only be 32 `onpagemasks`, since they are tracked trough the [`bitfield` parameter in the PTE](https://github.com/ComputationalRadiationPhysics/mallocMC/blob/dev/src/include/mallocMC/creationPolicies/Scatter_impl.hpp#L161-L173) (see Fig. 1)

### Figure 1: correctly working page with sane pagesize and chunksize
fullsegmentsbug_1

When the `pagesize` is very large and the allocated chunks very small (smaller than pagesize/1024), a **bug** would occur, that allows more than 32 `fullsegments`. Of course, these can never be used, since they are not found (no metadata for their `onpagemasks` in the PTE) (see Fig. 2)

### Figure 2: bug, when the chunksize is very small
fullsegmentsbug_2

After this patch, the `fullsegments` are calculated correctly and the `additional_chunks` as well.
### Figure 3: how the page should look like for *"too small"* chunksizes
fullsegmentsbug_3

## Additional Stuff
- calculation of `additional_chunks` needed to be restricted as a direct result of changing the formula for `fullsegments`, otherwise `additional_chunks` could get very large. This was put into a function, since it occurs at 3 places.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
@@ -263,6 +263,25 @@ namespace ScatterKernelDetail{
         }
       }
 
+
+      /**
+       * calcAdditionalChunks determines the number of chunks that are contained in the last segment of a hierarchical page
+       *
+       * The additional checks are necessary to ensure correct results for very large pages and small chunksizes
+       *
+       * @param fullsegments the number of segments that can be completely filled in a page. This may NEVER be bigger than 32!
+       * @param segmentsize the number of bytes that are contained in a completely filled segment (32 chunks)
+       * @param chunksize the chosen allocation size within the page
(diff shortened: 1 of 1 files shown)


ax3l:
fantastic pull request description :rocket:  :sparkles:

PR #80, opened 2015-01-22
Moved the onpagemasks to end of page

slizzered:
## TL;DR
`onpagemasks` is now always at the end of a page, so we save time when a page is freed.

## Full Text
When a page is reset, the area holding the `onpagetable` bitfields needs to be zero-filled. Doing this during the reset rather than during the allocation is thought to improve allocation times. However, the data layout of the page (see Fig. 1) leaves a lot of possible positions for the `onpagetables`, depending on the chunksize.
### Figure 1: Currently, `onpagetables` are located directly behind the data:
moved_onpagetables_1

Fig. 2 displays a worst-case scenario with small chunk sizes, where most of the page is empty and the `onpagetables` located near the beginning of the page.
### Figure 2: `onpagetables` is located near the beginning of the page:
moved_onpagetables_2

Since the zero-filling process takes place at the time the page is freed, it is impossible to know where the next allocation will place the `onpagetables`. Therefore, all possible positions for `onpagetable` need to be cleared (see Fig. 3). Benchmarks show that this is a major bottleneck when using large pages.
### Figure 3: considering the worst case, most of the page must be cleared:
moved_onpagetables_3

**This PullRequest** moves the `onpagetable` entries to the end of the page, without changing any other properties of the data (ordering of `onpagetable` entries stays the same). Fig. 4 depicts the page layout for the previous scenario with very small chunk sizes.
### Figure 4: even for small chunk sizes, `onpagetable` is located at the end:
moved_onpagetables_4

Since the location of `onpagetable` is now fixed, it is easy to determine how much of the page needs to be cleared. At most, 32 integers at the end of the page can be occupied by `onpagetables`. This might be even less for very small pages. The expected benefit is a much faster behaviour of `free()` in certain situations.

### Figure 5: only a small area needs to be cleared, since the position is fixed:
moved_onpagetables_5

## Additional Stuff
- This PR should be accepted AFTER #79 is merged to avoid possible heap corruptions (not sure, though).
- The calculation of the `onpagemask` offset occurs at 2 positions and was moved to a function.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
@@ -249,12 +249,6 @@ namespace ScatterKernelDetail{
        * @return pointer to the first address inside the page that holds metadata bitfields.
        */
       __device__ inline uint32* onPageMasksPosition(uint32 page, uint32 nMasks){
-        //uint32* x = (uint32*)(_page[page].data + pagesize - (int)sizeof(uint32)*nMasks);
-        //uint32* s = (uint32*)_page[page].data;
-        //uint32* e = (uint32*)(_page[page].data+pagesize);
-        //  printf("start %p end %p. diff=%lld  masks: %p diff_to_begin: %lld diff to end: %lld (pagesize is %d, there are %d masks)\n",
-        //          s, e, e-s, x, x-s, e-x, pagesize,nMasks);
-        //return x;
         return (uint32*)(_page[page].data + pagesize - (int)sizeof(uint32)*nMasks);
       }
 
--- a/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
@@ -132,6 +132,8 @@ namespace ScatterKernelDetail{
       //static const uint32 minChunkSize0 = pagesize/(32*32);
       static const uint32 minChunkSize1 = 0x10;
       static const uint32 HierarchyThreshold =  (pagesize - 2*sizeof(uint32))/33;
+      static const uint32 tmp_maxOPM = minChunkSize1 > HierarchyThreshold ? 0 : pagesize/(32*minChunkSize1 + 1);
+      static const uint32 maxOnPageMasks = 32 > tmp_maxOPM ? tmp_maxOPM : 32;
 
 #ifndef MALLOCMC_CP_SCATTER_HASHINGK
 #define MALLOCMC_CP_SCATTER_HASHINGK    static_cast<uint32>(HashingProperties::hashingK::value)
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
I love your descriptions! +2 :-)

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -132,6 +132,8 @@ namespace ScatterKernelDetail{
       //static const uint32 minChunkSize0 = pagesize/(32*32);
       static const uint32 minChunkSize1 = 0x10;
       static const uint32 HierarchyThreshold =  (pagesize - 2*sizeof(uint32))/33;
+      static const uint32 tmp_maxOPM = minChunkSize1 > HierarchyThreshold ? 0 : pagesize/(32*minChunkSize1 + 1);
```
psychocoderHPC: Why `+1`?
Should it be a integer round up like `(pagesize+(32*minChunkSize1-1)/(32*minChunkSize1)`?

slizzered: wow, good catch :+1: . I guess it was really late that night :sleeping: :crescent_moon: 
I think there are multiple errors in that line:
- for each full segment of the page, it takes `32*chunksize + sizeof(uint32)` byte. so the `+1` is flawed, since we count the number of bytes here:

`pagesize/(32*minChunkSize1 + sizeof(uint32))`
- for the (possible) **n** additional chunks, we need `n*chunksize+sizeof(32)`  byte, so there needs to be a round-up as well. It becomes:

`(pagesize + (32*minChunkSize1 + sizeof(uint32)) -1)/(32*minChunkSize1 + sizeof(uint32))`

Example1:
pagesize = 4096
minChunkSize = 16
-> space for 7 full segments (and their 7 onpagemasks), and 30 additional chunks (with 1 onpagemask)
`7*(32*16 + 4) + (30*16 + 4) = 4096`
The new formula gives **8** as a result, which is the necessary number of onpagemasks.

psychocoderHPC: Thanks for the fast reply.
Is it possible to create a `static const` variable for `(32*minChunkSize1 + sizeof(uint32))` to avoid mistakes?

slizzered: done, see next commit

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -239,6 +241,23 @@ namespace ScatterKernelDetail{
...
 
+
+      /**
+       * onPageMasksPosition returns a pointer to the beginning of the onpagemasks inside a page.
+       * @param page the page that holds the masks
+       * @param the number of hierarchical page tables (bitfields) that are used inside this mask.
+       * @return pointer to the first address inside the page that holds metadata bitfields.
+       */
+      __device__ inline uint32* onPageMasksPosition(uint32 page, uint32 nMasks){
+        //uint32* x = (uint32*)(_page[page].data + pagesize - (int)sizeof(uint32)*nMasks);
```
psychocoderHPC: Do we need this debug code?

slizzered: we don't, it was more for testing. will remove it tomorrow

psychocoderHPC:
Well done, I tested this pull last week and see very good performance with 2MiB  pages and 8KiB allocations.

Issue #28, opened 2014-01-09
usable copy of the heap in host memory


psychocoderHPC:
A copy of the heap memory should be useable on host memory.

IMPORTANT: all pointer addresses should be device pointer, that means on host any pointer address must translate to host memory address on each access.

PR #86, opened 2015-05-28
added host function to return heap pointer

slizzered:
This is a review for discussion (and potentially merging after a while)

In the current version, this adds a HeapInfo object to the allocator, which is used to track the pointer and the exact size of the heap.  

For future improvements, there should be a datastructure that holds multiple HeapInfo objects, since it is not guaranteed that there is only 1 heap. The idea is, that each HeapInfo object could hold information about how to access the objects in the memory-region (which allocator to use) and how the different regions are related (multiple HeapInfo objects must be in a certain order and belong to the same allocator).

See #28

Text form the commit message:
- use: std::vector<HeapInfo> mallocMC::getHeapLocations()
- uses a vector, since future allocators might have more than 1 heap
   (and those are not necessarily located one after another)
- currently, HeapInfo contains size and a pointer p
- the pointer is the device address (as it would be returned my
   cudaMalloc for example)
- the CUDA toolkit allocator will currently return NULL. Ideas very
   welcome. Problem: we have no pointer address for its internal heap
   (might be multiple heaps, who knows). Also, we don't know how objects
   are organized on the heap, so copying to host might be of limited
   value anyway.

### DIFF
--- a/src/include/mallocMC/mallocMC_hostclass.hpp
+++ b/src/include/mallocMC/mallocMC_hostclass.hpp
@@ -57,6 +57,11 @@ namespace mallocMC{
     static const bool providesAvailableSlots = T::CreationPolicy::providesAvailableSlots::value;
   };
 
+  class HeapInfo{
+    public:
+      void* p;
+      size_t size;
+  };
 
   /**
    * @brief "HostClass" that combines all policies to a useful allocator
@@ -96,6 +101,7 @@ namespace mallocMC{
     private:
       typedef boost::uint32_t uint32;
       void* pool;
+      HeapInfo heapInfos;
 
     public:
 
@@ -129,6 +135,8 @@ namespace mallocMC{
         pool = ReservePoolPolicy::setMemPool(size);
         boost::tie(pool,size) = AlignmentPolicy::alignPool(pool,size);
         void* h = CreationPolicy::initHeap(*this,pool,size);
+        heapInfos.p=pool;
+        heapInfos.size=size;
 
         /*
         * This is a workaround for a bug with getAvailSlotsPoly:
@@ -168,6 +176,13 @@ namespace mallocMC{
       MAMC_HOST MAMC_ACCELERATOR
       unsigned getAvailableSlots(size_t slotSize){
         return getAvailSlotsPoly(slotSize, boost::mpl::bool_<CreationPolicy::providesAvailableSlots::value>());
+      }
+
+      MAMC_HOST
+      std::vector<mallocMC::HeapInfo> getHeapLocations(){
+        std::vector<mallocMC::HeapInfo> v;
+        v.push_back(heapInfos);
+        return v;
       }
 
     private:
--- a/src/include/mallocMC/mallocMC_overwrites.hpp
+++ b/src/include/mallocMC/mallocMC_overwrites.hpp
@@ -103,6 +103,20 @@ void  free(void* p) __THROW                                                    \
 } /* end namespace mallocMC */
 
 
+/** Create the function getHeapLocations inside a namespace
+ *
+ * This returns a vector of type mallocMC::HeapInfo. The HeapInfo should at least
+ * contain the pointer to the heap (on device) and its size.
+ */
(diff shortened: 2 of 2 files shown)


psychocoderHPC:
:+1: 

We can define that the method `mallocMC::getHeapLocations()` is only callable direct after a heap is created.
For the cuda allocator we can run a kernel which allocate N small chunks of memory (16Byte might be the smalles size of an allocation) until the new return `NULL`. While we allocate memory we search the biggest and smallest pointer address. 
`size of heap=max(address)+16Byte-min(address)`
`begin_ptr=min(address)`

src/include/mallocMC/mallocMC_overwrites.hpp
```diff
@@ -103,6 +103,20 @@ void  free(void* p) __THROW                                                    \
...
 
+/** Create the function getHeapLocations inside a namespace
+ *
+ * This returns a vector of type mallocMC::HeapInfo. The HeapInfo should at least
+ * contain the pointer to the heap (on device) and its size.
+ */
+#define MALLOCMC_HEAPLOC()                                                     \
+namespace mallocMC{                                                            \
+MAMC_HOST                                                                      \
+std::vector<mallocMC::HeapInfo> getHeapLocations()                             \
```
psychocoderHPC: you  need to include vector in this file `#include <vector>`

psychocoderHPC: `#include <vector>` is missing in this file

src/include/mallocMC/mallocMC_hostclass.hpp
```diff
@@ -170,6 +178,13 @@ namespace mallocMC{
         return getAvailSlotsPoly(slotSize, boost::mpl::bool_<CreationPolicy::providesAvailableSlots::value>());
       }
 
+      MAMC_HOST
+      std::vector<mallocMC::HeapInfo> getHeapLocations(){
```
psychocoderHPC: `#include <vector>` is missing in this file

psychocoderHPC: You need to predefine `std::vector<HeapInfo>` in the top of the class else you get errors like this:

```
[ 16%] Building NVCC (Device) object build_picongpu/CMakeFiles/picongpu.dir//./picongpu_generated_main.cu.o
/home/widera/lib/mallocMC/include/mallocMC/mallocMC_hostclass.hpp(182): error: qualified name is not allowed

/home/widera/lib/mallocMC/include/mallocMC/mallocMC_hostclass.hpp(182): error: this declaration has no storage class or type specifier

/home/widera/lib/mallocMC/include/mallocMC/mallocMC_hostclass.hpp(181): error: attribute "host" does not apply here

/home/widera/lib/mallocMC/include/mallocMC/mallocMC_hostclass.hpp(182): error: expected a ";"

/home/widera/lib/mallocMC/include/mallocMC/mallocMC_hostclass.hpp(204): warning: parsing restarts here after previous syntax error

/bigdata/hplsim/scratch/widera/dev/src/picongpu/main.cu(74): error: class "mallocMC::Allocator<mallocMC::CreationPolicies::Scatter<ScatterConfig, mallocMC::CreationPolicies::ScatterConf::DefaultScatterHashingParams>, mallocMC::DistributionPolicies::Noop, mallocMC::OOMPolicies::ReturnNull, mallocMC::ReservePoolPolicies::SimpleCudaMalloc, mallocMC::AlignmentPolicies::Shrink<mallocMC::AlignmentPolicies::ShrinkConfig::DefaultShrinkConfig>>" has no member "getHeapLocations"
```

Define `typedef std::vector<HeapInfo> HeapInfoVector;` in the top of the class helps to avoid compile time errors. Errors are only triggert if `mallocMC:::getHeapLocations()` is used in the code.

psychocoderHPC:
I pushed https://github.com/slizzered/mallocMC/pull/1 to you pull request to fix compiler problems while using this pull request with [PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu).

PR #87, opened 2015-06-10
fix non-matching class/struct declaration

BenjaminW3:
The `BadAllocException` class is defined using `struct` but is forward declared using `class`.
Because `BadAllocException` does not have any private members I propose to change the forward declaration to `struct`.


slizzered:
:+1: Thanks for spotting this!

Issue #88, opened 2015-06-25
getAvailableSlotsHost is required even though the implementation may not provide it


BenjaminW3:
Currently the `getAvailableSlotsHost` function of a `CreationPolicy` is required irrespective of its availability.
In the [source](https://github.com/ComputationalRadiationPhysics/mallocMC/blob/84c8c20309633c98f69be83153eb897ce2d83110/src/include/mallocMC/mallocMC_hostclass.hpp#L144-155) the `CreationPolicy::providesAvailableSlots::value` flag is checked in a runtime `if` so the function is required for all CreationPolicies even for `OldMalloc`.
This has to be changed to a compile time construct.

BenjaminW3:
Something along the lines

```C++
template<typename T>
struct CallGetAvailableSlotsHostIfAvail
{
  void operator()()
  {}
};
template<>
struct CallGetAvailableSlotsHostIfAvail<boost::mpl::bool_<true>>
{
  void operator()()
  {
    CreationPolicy::getAvailableSlotsHost(1024,*this); //actual slot size does not matter
  }
};
```

called with

```C++
CallGetAvailableSlotsHostIfAvail<CreationPolicy::providesAvailableSlots>()();
```

should do it.

PR #89, opened 2015-06-26
fix unconditional usage of getAvailableSlotsHost

BenjaminW3:
Currently the `getAvailableSlotsHost` function of a `CreationPolicy` is required irrespective of its availability.
In the [source](https://github.com/ComputationalRadiationPhysics/mallocMC/blob/84c8c20309633c98f69be83153eb897ce2d83110/src/include/mallocMC/mallocMC_hostclass.hpp#L144-155) the `CreationPolicy::providesAvailableSlots::value` flag is checked in a runtime `if` so the function is required for all CreationPolicies even for `OldMalloc`.
This has been changed to a compile time construct.

See issue #88.

### DIFF
--- a/src/include/mallocMC/mallocMC_hostclass.hpp
+++ b/src/include/mallocMC/mallocMC_hostclass.hpp
@@ -64,6 +64,42 @@ namespace mallocMC{
       size_t size;
   };
 
+  namespace detail{
+
+    /**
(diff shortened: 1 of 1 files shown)


ax3l:
@slizzered can you pls take a look? :)

slizzered:
I like the solution a lot, thanks @BenjaminW3 !

PR #91, opened 2015-07-06
Define __THROW for systems without Glibc

BenjaminW3:
__THROW is defined in Glibc so it is not available on all platforms.
The safest way to fix this is to define it empty for platforms where it is not available.
Maybe it is also safe to remove it completely.


slizzered:
I don't have the microsoft compiler at my disposal, so I can't reliably test the case that this is probably intended for. However, defining it like this appears to be quite harmless and safe.

BenjaminW3:
With this definition it works with MSVC (without it did not).

PR #93, opened 2015-07-29
Close #92: Update CMake Version to 2.8.12.2+

### LINKED ISSUE #92: CMake: 2.8.12.2+
Since the default CMake [module](https://github.com/ComputationalRadiationPhysics/cmake-modules) does now require cmake 2.8.12.2+, one could push the cmake versions for the project itself (`cmake_minimum_required`) and the documentation of required software itself to the same version to avoid confusion for users.

(Aka "I installed in successfully but my project does not work with it..." -> let's force a CMake version that will hold the full way to a project.)

ax3l:
Require CMake 2.8.12.2+ for the whole project to ensure the full build chain until a project that is using the FindmallocMC.cmake module will pass. Closes #92.

Dependency increased in
  https://github.com/ComputationalRadiationPhysics/cmake-modules/pull/8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,9 @@ script:
   - make examples
 
 before_script:
+  - sudo add-apt-repository --yes ppa:smspillaz/cmake-2.8.12
   - sudo apt-get update -qq
+  - sudo apt-get install -q -y cmake-data cmake
   - sudo apt-get install -qq build-essential
   - sudo apt-get install -qq gcc-4.4 g++-4.4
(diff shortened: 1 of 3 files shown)


ax3l:
of course travis, you are running cmake `2.8.7`...

slizzered:
just for documentation: the link and headline of the page says 2.8.12, but the ppa is actually 2.8.12.2 :smile: 

Thanks for fixing this!

ax3l:
yes it is, a bit hidden in the module naming itself when one scrolls to the `Overview of published packages` section.

PR #95, opened 2015-08-27
Wshadow Compliance

ax3l:
Due to the effort to confirm with `-Wshadow` in https://github.com/ComputationalRadiationPhysics/picongpu/issues/1042 , mallocMC should also fulfill this requirement.

This pull adds the flag to GCC & ICC and adds `-Werror` for compliance enforcement during the travis tests.

GCC:
 - Wshadow
 - Werror (used in compile suite)

ICC:
 - Wshadow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
 
 script:
   - mkdir build_tmp && cd build_tmp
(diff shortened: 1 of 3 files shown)


slizzered:
Thanks for fixing the Werror situation in the travis build (you were 5min faster than me ;) )

Issue #65, opened 2014-07-31
Boost Static Assert & Pre 7.0 CUDA


slizzered:
This is the reminder for a work around for versions of CUDA before `7.0` and gcc before `4.8.2`, see #97
## 

Test, if there is a compiler bug related to -Wsign-compare:

The code

``` c++
uint32 x = ...;
BOOST_STATIC_ASSERT(x > 0);
```

produces a compiler warning about comparison between signed and unsigned.

None of the following does fix the warning on my test system:

``` c++
BOOST_STATIC_ASSERT(x > 0U);
BOOST_STATIC_ASSERT(x > (uint32)0);
BOOST_STATIC_ASSERT(x > static_cast<unsigned>(0));
```

However,  this works (strangely enough):

``` c++
BOOST_STATIC_ASSERT(static_cast<uint32>(x) > 0);
```

To be found in mallocMC:
- https://github.com/ComputationalRadiationPhysics/mallocMC/blob/dev/src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp#L80
- https://github.com/ComputationalRadiationPhysics/mallocMC/blob/dev/src/include/mallocMC/alignmentPolicies/Shrink_impl.hpp#L75-L77

Related: 
https://github.com/ComputationalRadiationPhysics/mallocMC/pull/64

ax3l:
Isn't it normal that
```c++
uint32 x = ...;
BOOST_STATIC_ASSERT(x > 0);
``` produces a warning? `0` is an `int`, `0u` is an unsigned.

slizzered:
Yes it is normal. The problem is that `BOOST_STATIC_ASSERT(x > 0U);` and the like don't work as I would expect

ax3l:
ah, I over read the `None of the following  ...` !

ax3l:
Sounds like a boost with nvcc problem to me.

Can you cross check with CUDA 7.5 and the latest boost release and open an issue in both issue trackers?

slizzered:
I can no longer produce the warning (directly in the mentioned code segment) on my home machine (CUDA 7.0, gcc 5.2, Boost 1.58), which indicates that the bug has been resolved in the meantime.

That lead me to do some more testing:

| CUDA   |  Boost  | gcc   | status |
| ------ | --------| ------| ------ |
| 6.0    | 1.55.0  | 4.8.2 | :white_check_mark:     |
[...]

PR #97, opened 2015-08-28
Added TODO to boost_static_assert cast warning

slizzered:
- Since the warning (and the cast necessary to suppress it) are solved
   in newer versions of nvcc/gcc, one bright day in the future might
   permit us to remove that dreadful cast again.
 - see #65


ax3l:
thank you! :sparkles:

PR #100, opened 2015-09-18
removed typedef uint32_richtig_huebsch

slizzered:
- replaced by `uint32_t` from `<boost/cstdint.hpp>`

### DIFF
--- a/src/include/mallocMC/mallocMC_utils.hpp
+++ b/src/include/mallocMC/mallocMC_utils.hpp
@@ -101,7 +101,7 @@ namespace CUDA
 
 
 #define warp_serial                                    \
-  for (uint32_t __mask = __ballot(1),                    \
+  for (unsigned int __mask = __ballot(1),                    \
             __num = __popc(__mask),                    \
             __lanemask = mallocMC::lanemask_lt(),      \
             __local_id = __popc(__lanemask & __mask),  \
--- a/src/include/mallocMC/mallocMC_utils.hpp
+++ b/src/include/mallocMC/mallocMC_utils.hpp
@@ -101,7 +101,7 @@ namespace CUDA
 
 
 #define warp_serial                                    \
-  for (unsigned int __mask = __ballot(1),                    \
+  for (unsigned int __mask = __ballot(1),              \
             __num = __popc(__mask),                    \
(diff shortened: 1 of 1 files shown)


src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -100,7 +101,7 @@ namespace CUDA
 
 
 #define warp_serial                                    \
-  for (uint32_richtig_huebsch __mask = __ballot(1),                    \
+  for (uint32_t __mask = __ballot(1),                    \
```
ax3l: I think the header signatures of these CUDA intrinsics are unfortunately [unsigned int](http://docs.nvidia.com/cuda/cuda-c-programming-guide/#warp-vote-functions).

I would therefore either write this explicitly or keep a typedef (with an other name) but not assuming `uint32_t`

ax3l:
I would pass this to @psychocoderHPC if that is ok :)

ax3l:
why do you need boost headers and not simple `<cstdint>` from the std libs?

src/include/mallocMC/mallocMC_utils.hpp
slizzered: wow, well spotted!
I will use the type directly, since this is the only line where this problem occurs.

slizzered:
The rest of the project already uses `<boost/cstdint.hpp>`, so it was used for consistency.

AFAIK, `<cstdint>` is not available on Windows unless you use C++11, which was explicity not targeted at the time. So the Boost solution is mostly about compatibility between platforms and compiler versions.

ax3l:
oh true, I didn't know [<cstdint>](http://www.cplusplus.com/reference/cstdint/) just became part of the stdlibs outside the *nix world in C++11

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -100,7 +101,7 @@ namespace CUDA
 
 
 #define warp_serial                                    \
-  for (uint32_richtig_huebsch __mask = __ballot(1),                    \
+  for (unsigned int __mask = __ballot(1),                    \
```
psychocoderHPC: the backslash `\` is not aligned

PR #105, opened 2015-10-02
Fix #104 `uint32_t` cstdint

### LINKED ISSUE #104: utils: uint32_t undefined
Using
- CUDA 6.5
- cmake 3.0.1
- gcc 4.6.2
- boost 1.54.0

on mallocMC `2.2.0` in PIConGPU, I get unknown `uint32_t` errors during compile from utils:
```
mallocMC/src/include/mallocMC/mallocMC_utils.hpp(133): error:
  identifier "uint32_t" is undefined
[...]
```

Seems to be a regression to #100.

ax3l:
Fix #104: To select the `uint32_t` from `<boost/cstdint.hpp>` without ambigouity, one needs to select it from it's `boost::` scope.

Fixes the update from #100.

### DIFF
--- a/src/include/mallocMC/mallocMC_utils.hpp
+++ b/src/include/mallocMC/mallocMC_utils.hpp
@@ -130,70 +130,70 @@ namespace mallocMC
   typedef mallocMC::__PointerEquivalent<sizeof(char*)>::type PointerEquivalent;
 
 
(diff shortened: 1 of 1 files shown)


slizzered:
thanks for the fix!

PR #108, opened 2015-10-08
Rename some shadowed variables in C++11 mode

Flamefire:
It seems like a,b,c are used in cuda(?) functions in C++11.

This renames them to arA, arB and arC

### DIFF
--- a/examples/mallocMC_example01.cu
+++ b/examples/mallocMC_example01.cu
@@ -55,28 +55,28 @@ int main()
 }
 
 
-__device__ int** a;
-__device__ int** b;
-__device__ int** c;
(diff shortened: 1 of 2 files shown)


slizzered:
I was not able to reproduce the problem locally (CUDA 7.5), but renaming sounds like a straightforward idea (and there are probably more systems than yours that experience the problem).

Could you rebase the PR so that I can merge it, please?

And thanks for the effort of testing it!

Flamefire:
I think I had it with CUDA 7. So it seems it is fixed in CUDA 7.5.
Rebased.

slizzered:
thanks!

PR #107, opened 2015-10-08
Fix crash in getAvailableSlots

Flamefire:
Fixes #106 

This makes getAvailableSlots a host only function (most common case) and adds a function getAvailableSlotsAccelerator for device side calls. It allows also the removal of the workaround in initHeap

PR #109, opened 2015-10-08
Use BOOST_STATIC_CONSTEXPR where possible

Flamefire:
This uses BOOST_STATIC_CONSTEXPR instead of `static const` for C++11 compatibility although it seems to work without that too. But to avoid future problems I still added that.

Merge *AFTER* #107


ax3l:
I would propose to hold this merge to put it in the `2.3.0` release.

ax3l:
@Flamefire needs rebase.

Flamefire:
Rebased anyway if you like to include it

ax3l:
@Flamefire needs rebase :)

Flamefire:
Done

PR #112, opened 2015-10-15
Fixed problems leading to uninitialized pointers

### LINKED ISSUE #110: Missing error handling leading to undefined behaviour
The `MALLOCMC_CUDA_CHECKED_CALL` macro uses `CUDA::checkError` to check for error codes which throws an error in case of non-zero values. But this happens only in Debug mode.

Therefore when the common `SimpleCudaMalloc` policy is used and `setMemPool` is called with an invalid/to high size or the `cudaMalloc` call fails for any reason, an uninitialized pointer is returned.
This also holds true for e.g. sizes of 0 in which case cudaMalloc does not modify the pointer but returns a zero-error code indicating success. See the code:

    static void* setMemPool(size_t memsize){
      void* pool;
      MALLOCMC_CUDA_CHECKED_CALL(cudaMalloc(&pool, memsize));
      return pool;
    }

Even setting this pointer to zero and not throwing an error would not solve the problem, as the result of the call is never checked and implicitly assumed valid in all following usages.

As the minimum action to counter this I propose to enable the exception throwing also in Release mode and initialize the pointer to NULL before passing it to cudaMalloc

slizzered:
- CUDA::checkError now throws always
 - pointer in SimpleCudaMalloc::setMemPool is initialized
 - fixes #110

### DIFF
--- a/src/include/mallocMC/mallocMC_utils.hpp
+++ b/src/include/mallocMC/mallocMC_utils.hpp
@@ -75,10 +75,8 @@ namespace CUDA
 
   inline void checkError(cudaError errorValue, const char* file, int line)
   {
-#ifdef _DEBUG
     if (errorValue != cudaSuccess)
       throw CUDA::error(errorValue, file, line);
-#endif
   }
 
   inline void checkError(const char* file, int line)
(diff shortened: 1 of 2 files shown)


ax3l:
@psychocoderHPC pls review and merge if appropriate

PR #119, opened 2015-12-08
Updating to Travis-CI Trusty Beta

ax3l:
GCC 4.8.2 + CUDA 5.5 + Native CMake 2.8.12.2 Images

I am not sure if it is relevant, since for some reasons travis-ci just worked again yesterday.

Maybe it is not too bad to have an old CUDA and GCC version tested automatically. Nevertheless, if we should need a newer version at some point, here is a new image without PPA's.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: cpp
 
+sudo: required
+
+dist: trusty
+
 compiler:
   - gcc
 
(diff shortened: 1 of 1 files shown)


ax3l:
@slizzered merge me: you now need `multiverse` to get CUDA installed (5.5 in trusty).

precise (cuda 4.0) would need precise-backports, so let's go to trusty (still super old dist).

slizzered:
finally a (relatively...) new Ubuntu :)
thanks a lot!

PR #121, opened 2016-01-22
Fix: Includes from C headers

ax3l:
Use correct `<c...>` includes in C++.

Only thing I was not sure about because it seems to be non-std:
```
./src/include/mallocMC/mallocMC_utils.hpp :

```

@slizzered

### DIFF
--- a/examples/mallocMC_example01.cu
+++ b/examples/mallocMC_example01.cu
@@ -27,7 +27,7 @@
 */
 
 #include <iostream>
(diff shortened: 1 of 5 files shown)


slizzered:
@ax3l want to try it again? (just merged #119 )

ax3l:
yes, rebased :)

slizzered:
cool, thanks!

PR #126, opened 2016-06-13
Check heap pointer in Scatter creation policy

### LINKED ISSUE #115: Scatter policy does not check heap for NULL pointer
As described in #111 and #110 it is possible that a NULL pointer is passed to the scatter policy with a size != 0. That policy would then use the pointer, assuming it to be valid without any further checks eventually leading to a crash.

slizzered:
fix #115

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
@@ -4,8 +4,8 @@
 
   Copyright (C) 2012 Institute for Computer Graphics and Vision,
                      Graz University of Technology
-  Copyright (C) 2014 Institute of Radiation Physics,
-                     Helmholtz-Zentrum Dresden - Rossendorf
+  Copyright (C) 2014-2016 Institute of Radiation Physics,
+                          Helmholtz-Zentrum Dresden - Rossendorf
 
   Author(s):  Markus Steinberger - steinberger ( at ) icg.tugraz.at
               Rene Widera - r.widera ( at ) hzdr.de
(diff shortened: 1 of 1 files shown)


slizzered:
don't merge yet. there appear to be some warnings

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -817,6 +818,13 @@ namespace ScatterKernelDetail{
 
       template < typename T_DeviceAllocator >
       static void* initHeap( T_DeviceAllocator* heap, void* pool, size_t memsize){
+        if( pool == NULL && memsize != 0 )
+        {
+          throw std::invalid_argument(
+            "Scatter policy cannot use NULL for non-empty memory pools.\
```
ax3l: Should be newlines as
```C++
    "something. "
    "New sentence."

slizzered: Not sure I understand your point. The constructor of `invalid_argument` only takes a single string literal as input. The sentences were only split into 2 lines to comply with the 80 character limit per line.

I believe the final output should be on a single line to make clear that the sentences together and originate at the same place.

ax3l: the point is: the syntax I gave you is the correct way to split *a single* string argument.

slizzered: Wow, I was actually not aware of the correct way to split a string in C++. Thanks for your input! :sparkling_heart:

ax3l: :sparkles:

slizzered:
ok, can be merged now :)

ax3l:
cool, thanks!

PR #128, opened 2016-06-14
Fix to travis file: -Werror

slizzered:
- -Werror was not working correctly (warnings did not produce errors in
   CI). Hopefully this is fixed now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,12 @@ compiler:
 env:
   global:
     - INSTALL_DIR=~/mylibs
+    - CMAKE_FLAGS="-DCMAKE_CXX_FLAGS=\"-Werror\" -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR"
+  # - CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=$INSTALL_DIR"
 
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,10 +10,12 @@ compiler:
 env:
   global:
     - INSTALL_DIR=~/mylibs
+    - CMAKE_FLAGS="-DCMAKE_CXX_FLAGS=\"-Werror\" -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR"
```
ax3l: why not simply export CXXFLAGDS="-Werror" at this point and keep install prefix below?

ax3l: example: https://github.com/pngwriter/pngwriter/blob/0.5.6/.travis.yml#L7-L9

slizzered: I had taken the idea from https://github.com/ComputationalRadiationPhysics/cupla/blob/dev/.travis.yml#L41, since the old way didn't work ( `CXX_FLAGS="-Werror" cmake` ), probably because the current value of CXX_FLAGS was not integrated into the Makefile (as I blindly assumed).
Storing it as a global env-var will probably work (fix incoming!)

PR #138, opened 2017-01-25
add destructor and documentaion

psychocoderHPC:
- destructor: free allocated memory
- add constructor documentation


psychocoderHPC:
`finalizeHeap` is not destroying the memory it is only a clear function. Therefore I can't use this function for the constructor

PR #139, opened 2017-01-25
fix `finalizeHeap`

psychocoderHPC:
- device handle is destroyed and can used with invalid memory
- revert #135 because `finalizeHeap()` only clear the memory


psychocoderHPC:
Fail again, `finalizeHeap()` is confusing me, it calls some clear function and also some destroy functions.
I will fix it in a follow up.

PR #140, opened 2017-01-25
self consistent allocator

psychocoderHPC:
With `finalizeHeap` it is possible that the user can create an invalid allocator and access invalid memory.
Wrong usage can be avoided by removing this method and forbid the copy constructor.

- allocator.hpp
  - remove `finalizeHeap()`
  - add private `free()` method
  - forbid copy constructor
  - add destructor
- update `Usage.md`
- creationPolicies: remove `finalizeHeap`
  - it is not possible to implement a `reset` function for each `creationPolicy` e.g. `OldMalloc`
- update examples and tests

### DIFF
--- a/Usage.md
+++ b/Usage.md
@@ -13,7 +13,7 @@ There is one header file that will include *all* necessary files:
 Step 2a: choose policies
 -----------------------
 
-Each instance of a policy based allocator is composed through 5 **policies**. Each policy is expressed as a **policy class**. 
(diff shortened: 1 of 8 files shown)


ax3l:
@slizzered just in case you are interested, any feedback on that?

ax3l:
from my point of view reasonable, will merge for now into `dev` :)

PR #144, opened 2017-11-01
support for CUDA9 `__ballot_sync`

psychocoderHPC:
- use `ballot_sync` if CUDA9 is used
- copy code of precompiler function `warp_serial` to the used function

This PR will remove the warning: `mallocMC/src/include/mallocMC/creationPolicies/Scatter_impl.hpp(620): warning: function "__ballot"`

PR #145, opened 2018-05-16
Fix V100 Active Lanes in Warp

ax3l:
Fix for V100 and CUDA 9.0 ballot functions, introduced in #144.
Properly use the currently active mask of lanes in a warp instead of a constant.

We have seen deadlocks in [PIConGPU/PMacc](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2600) with when using the `FULL_MASK` (`0xFFFFFFFF`). This can cause deadlocks e.g. in branched threads which are properly described with `__activemask()` instead.

https://devblogs.nvidia.com/using-cuda-warp-level-primitives/


ax3l:
@psychocoderHPC ping :)

PR #150, opened 2019-01-04
fix illegal memory access

psychocoderHPC:
fix #149

The number of warps per multiprocessor depends on the architecture.
In some places, the warp id was used to access block shared memory with a fixed size of 32.
Since sm_20 the number of warps per multiprocessor is 64 which can create an out of memory access.

- add helper methods for:
  - MaxThreadsPerBlock
  - WarpSize
  - warpid_withinblock()
- fix collective warp aggregations

### DIFF
--- a/src/include/mallocMC/mallocMC_utils.hpp
+++ b/src/include/mallocMC/mallocMC_utils.hpp
@@ -122,12 +122,24 @@ namespace mallocMC
     return mylaneid;
   }
 
+  /** warp index within a multiprocessor
+   *
+   * Index of the warp within the multiprocessor at the moment of the query.
+   * The result is volatile and can be different with each query.
+   *
+   * @return current index of the warp
+   */
   MAMC_ACCELERATOR inline boost::uint32_t warpid()
   {
     boost::uint32_t mywarpid;
     asm("mov.u32 %0, %%warpid;" : "=r" (mywarpid));
     return mywarpid;
   }
+
+  /** maximum number of warps on a multiprocessor
+   *
+   * @return maximum number of warps on a multiprocessor
+   */
   MAMC_ACCELERATOR inline boost::uint32_t nwarpid()
   {
     boost::uint32_t mynwarpid;
@@ -192,23 +204,23 @@ namespace mallocMC
    */
   struct MaxThreadsPerBlock
   {
-	// valid for sm_2.X - sm_7.5
-    BOOST_STATIC_ASSERT uint32_t value = 1024;
+    // valid for sm_2.X - sm_7.5
+    BOOST_STATIC_CONSTEXPR uint32_t value = 1024;
   };
 
-  /** the maximal number threads per block
+  /** number of threads within a warp
    *
    * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
    */
   struct WarpSize
   {
-	// valid for sm_2.X - sm_7.5
-    BOOST_STATIC_ASSERT uint32_t value = 32;
+    // valid for sm_2.X - sm_7.5
+    BOOST_STATIC_CONSTEXPR uint32_t value = 32;
   };
 
-  /** warp id within a block
+  /** warp id within a cuda block
    *
-   * The id is constant over the livetime of the thread.
+   * The id is constant over the lifetime of the thread.
    * The id is not equal to warpid().
    *
    * @return warp id within the block
@@ -218,7 +230,7 @@ namespace mallocMC
     return (
       threadIdx.z * blockDim.y * blockDim.x +
       threadIdx.y * blockDim.x +
-	  threadIdx.x
+      threadIdx.x
     ) / WarpSize::value;
   }
 }
--- a/src/include/mallocMC/mallocMC_utils.hpp
+++ b/src/include/mallocMC/mallocMC_utils.hpp
@@ -186,4 +186,39 @@ namespace mallocMC
   template<class T>
   MAMC_HOST MAMC_ACCELERATOR inline T divup(T a, T b) { return (a + b - 1)/b; }
 
+  /** the maximal number threads per block
+   *
+   * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
+   */
+  struct MaxThreadsPerBlock
+  {
+	// valid for sm_2.X - sm_7.5
+    BOOST_STATIC_ASSERT uint32_t value = 1024;
+  };
+
+  /** the maximal number threads per block
+   *
+   * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
+   */
+  struct WarpSize
+  {
+	// valid for sm_2.X - sm_7.5
+    BOOST_STATIC_ASSERT uint32_t value = 32;
+  };
+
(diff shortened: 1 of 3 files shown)


src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
   template<class T>
   MAMC_HOST MAMC_ACCELERATOR inline T divup(T a, T b) { return (a + b - 1)/b; }
 
+  /** the maximal number threads per block
+   *
+   * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
+   */
+  struct MaxThreadsPerBlock
+  {
+	// valid for sm_2.X - sm_7.5
```
ax3l: tabs

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
+    BOOST_STATIC_ASSERT uint32_t value = 1024;
+  };
+
+  /** the maximal number threads per block
```
ax3l: copy paste in docs title :)

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
+   *
+   * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
+   */
+  struct WarpSize
+  {
+	// valid for sm_2.X - sm_7.5
```
ax3l: tabs

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
+    BOOST_STATIC_ASSERT uint32_t value = 32;
+  };
+
+  /** warp id within a block
+   *
+   * The id is constant over the livetime of the thread.
```
ax3l: needs more details please:
the warpid you calculate here is: unique for the warps in a block over the lifetime of a whole block, right?
Because it could also be unique only for the active number of threads, etc.

typo: lifetime

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
+   * The id is not equal to warpid().
+   *
+   * @return warp id within the block
```
ax3l: in `warpid()` we should add a comment that it should not be used besides for diagnostics.
Maybe we want to implement the [`%ctaid and %tid` getter](https://github.com/ComputationalRadiationPhysics/mallocMC/pull/149#issuecomment-451088827) as described by the docs.

Can you please `grep` the code after your change to check no `warpid` is used anywhere else?

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
+   */
+  MAMC_ACCELERATOR inline boost::uint32_t warpid_withinblock()
+  {
+    return (
+      threadIdx.z * blockDim.y * blockDim.x +
```
ax3l: that implementation is likely not true if any of he block dims is not exactly a multiple of the warp size.

ax3l:
```
For this reason, %ctaid and %tid should be used to compute a virtual warp index if such a value is needed in kernel code
```

ax3l: btw, didn't know that `%warpid` is also quite expensive (besides being wrong for our case): https://devtalk.nvidia.com/default/topic/1011523/cuda-programming-and-performance/how-costly-is-the-s2r-instruction-reading-a-special-register-/post/5165296/#5165296

src/include/mallocMC/mallocMC_utils.hpp
psychocoderHPC: > the warpid you calculate here is: unique for the warps in a block over the lifetime of a whole block, right?

Yes, but since a thread can only have a warpid during the lifetime, the doc is correct. As long as the thread exists it has a corresponding warpid. It do not matter if the thread is active or inactive.

src/include/mallocMC/mallocMC_utils.hpp
psychocoderHPC: > in warpid() we should add a comment that it should not be used besides for diagnostics.

We can extend the documentation but the point with the diagnostic is not correct e.g. `warpid` is still used for a hashing function where it makes sense and it is not critical. 

> Maybe we want to implement the %ctaid and %tid getter as described by the docs.

I do not understand how this helps because `%%tid == threadIdx`

src/include/mallocMC/mallocMC_utils.hpp
psychocoderHPC: > that implementation is likely not true if any of the block dims is not exactly a multiple of the warp size.

It is not required that it is a multiple. It gives the linear thread idx and than device it by the warp size.
Could you please explain why it should be wrong?

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
   MAMC_HOST MAMC_ACCELERATOR inline T divup(T a, T b) { return (a + b - 1)/b; }
 
+  /** the maximal number threads per block
+   *
+   * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
+   */
+  struct MaxThreadsPerBlock
+  {
+	// valid for sm_2.X - sm_7.5
+    BOOST_STATIC_ASSERT uint32_t value = 1024;
```
matthias-springer: I am getting a compile error here. Shouldn't there be parentheses around any static assert? Or what does BOOST_STATIC_ASSERT do here?

psychocoderHPC: ohh it is a copy past mistake it must be `BOOST_STATIC_CONSTEXPR`

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -186,4 +186,39 @@ namespace mallocMC
...
+  };
+
+  /** the maximal number threads per block
+   *
+   * https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities
+   */
+  struct WarpSize
+  {
+	// valid for sm_2.X - sm_7.5
+    BOOST_STATIC_ASSERT uint32_t value = 32;
```
psychocoderHPC: must be `BOOST_STATIC_CONSTEXPR`

src/include/mallocMC/mallocMC_utils.hpp
ax3l: offline discussion: the question is just about the strategy of thread-linearization to warps at runtime. do we have a reference that this is the same mapping during thread scheduling?

psychocoderHPC:
@matthias-springer I fixed my copy past issue. Could you please test if this PR solved the issue for you.

matthias-springer:
@psychocoderHPC It works! Thank you!

PR #164, opened 2019-09-24
Choose the value for the -arch nvcc flag depending on CUDA version

### LINKED ISSUE #162: Hardcoded flag for compute capability 2.0
[This line](https://github.com/ComputationalRadiationPhysics/mallocMC/blob/e2533d14101c9fa7af3d11b9d02277591e06d8e4/CMakeLists.txt#L25) has a hardcoded flag for compute capability 2.0, which is not supported since CUDA 9. So examples do not compile with the compiler stack used by Alpaka and PIConGPU.
The cmake file generally has room for improvement, but this is the most immediate issue.

sbastrakov:
The original code has hardcoded 2.0, not supported since CUDA 9.
Closes #162.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,12 @@ endif()
 # CUDA
 ###############################################################################
 find_package(CUDA REQUIRED)
-set(CUDA_NVCC_FLAGS "-arch=sm_20;-use_fast_math;")
+if(CUDA_VERSION VERSION_LESS 9.0)
+    set(COMPUTE_CAPABILITY "20")
+else()
+    set(COMPUTE_CAPABILITY "30")
+endif()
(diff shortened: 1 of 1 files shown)


sbastrakov:
Note: the documentation and examples still have 2.0 hardcoded, while with this PR that would be 2.0 for CUDA < 9 and 3.0 for CUDA >= 9. After an offline discussion with @psychocoderHPC we decided to leave it as is for now.

CMakeLists.txt
```diff
@@ -22,7 +22,12 @@ endif()
 # CUDA
 ###############################################################################
 find_package(CUDA REQUIRED)
-set(CUDA_NVCC_FLAGS "-arch=sm_20;-use_fast_math;")
+if(CUDA_VERSION VERSION_LESS 9.0)
```
ax3l: Please add a `if(NOT DEFINED COMPUTE_CAPABILITY)` around this, so one can configure it externally.

PR #169, opened 2020-05-06
Cleaning & C++14 Migration

bernhardmgruber:
@bussmann has asked my to read a bit into mallocMC, so I spent the afternoon today digging through the code. Since I also love refactoring, I made a few changes on the way.

* requiring C++14
* using cstdint instead of boost/cstdint.hpp
* using std::tuple instead of boost::tuple
* using nullptr
* using static_assert
* using constexpr
* adding const and static where appropriate
* removed a few empty lines
* replaced std::endl by \n where flush was probably not intended
* replaced usage of boost::mpl by static constexpr members
* dropped dependency on boost

All examples and the verifyHeap target build, run and show some meaningful output. But I do not understand the details to be honest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 project(mallocMC)
 cmake_minimum_required(VERSION 2.8.12.2)
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 # helper for libs and packages
 set(CMAKE_PREFIX_PATH "/usr/lib/x86_64-linux-gnu/"
     "$ENV{CUDA_ROOT}" "$ENV{BOOST_ROOT}")
@@ -56,7 +59,7 @@ set(LIBS ${LIBS} ${Boost_LIBRARIES})
 # nvcc + boost 1.55 work around
 if(Boost_VERSION EQUAL 105500)
   set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} \"-DBOOST_NOINLINE=__attribute__((noinline))\" ")
-endif(Boost_VERSION EQUAL 105500)
+endif()
 
 
 ################################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,13 +24,9 @@ endif()
 ###############################################################################
 # CUDA
 ###############################################################################
-find_package(CUDA REQUIRED)
+find_package(CUDA 9.0 REQUIRED)
 if(NOT DEFINED COMPUTE_CAPABILITY)
-    if(CUDA_VERSION VERSION_LESS 9.0)
-        set(COMPUTE_CAPABILITY "20")
-    else()
-        set(COMPUTE_CAPABILITY "30")
-    endif()
+    set(COMPUTE_CAPABILITY "30")
 endif()
 set(CUDA_NVCC_FLAGS "-arch=sm_${COMPUTE_CAPABILITY};-use_fast_math;")
 set(CUDA_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,19 +49,6 @@ if(NOT ${CUDA_OPTIMIZATION_TYPE} STREQUAL  "unset")
 endif()
 
 
-###############################################################################
-# Boost
-###############################################################################
-find_package(Boost 1.48.0 REQUIRED)
-include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
-set(LIBS ${LIBS} ${Boost_LIBRARIES})
-
-# nvcc + boost 1.55 work around
-if(Boost_VERSION EQUAL 105500)
-  set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} \"-DBOOST_NOINLINE=__attribute__((noinline))\" ")
-endif()
-
-
 ################################################################################
 # Warnings
 ################################################################################
@@ -117,8 +104,3 @@ cuda_add_executable(mallocMC_Example03
 cuda_add_executable(VerifyHeap
                     EXCLUDE_FROM_ALL
                     tests/verify_heap.cu )
-
-target_link_libraries(mallocMC_Example01 ${LIBS})
-target_link_libraries(mallocMC_Example02 ${LIBS})
-target_link_libraries(mallocMC_Example03 ${LIBS})
-target_link_libraries(VerifyHeap ${LIBS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,12 @@
-project(mallocMC)
-cmake_minimum_required(VERSION 2.8.12.2)
+project(mallocMC LANGUAGES CUDA CXX)
+cmake_minimum_required(VERSION 3.8)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # helper for libs and packages
+set(CMAKE_CUDA_STANDARD 14)
+set(CMAKE_CUDA_STANDARD_REQUIRED ON)
 set(CMAKE_PREFIX_PATH "/usr/lib/x86_64-linux-gnu/"
     "$ENV{CUDA_ROOT}" "$ENV{BOOST_ROOT}")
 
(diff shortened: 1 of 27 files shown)


src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -182,7 +179,7 @@ namespace ScatterKernelDetail{
        * The page struct is used to access the data on the page more efficiently
        * and to clear the area on the page, which might hold bitsfields later one
        */
-      struct PAGE
+      struct PAGE // TODO(bgruber): why is this in ALLCAPS?
```
psychocoderHPC, line 182: I think it was Caps in the original scatter implementation and is copy pasted.

I would suggest to change it to `Page` to fit our [coding style](https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md), note: The rest of the code also do not fit the coding style because the code is older than our style guide.

ax3l:
Awesome, thanks a lot!
Feel free to bump the CI to require CUDA 9.2+. Anything else is considered ancient and unusable for proper C++14.

ax3l:
Opened a CUDA 9.1.85 PR in #170, maybe this works already for you.

src/include/mallocMC/alignmentPolicies/Shrink_impl.hpp
```diff
@@ -70,18 +68,18 @@ namespace Shrink2NS{
...
-#define MALLOCMC_AP_SHRINK_DATAALIGNMENT Properties::dataAlignment::value
+#define MALLOCMC_AP_SHRINK_DATAALIGNMENT (Properties::dataAlignment)
 #endif
-    BOOST_STATIC_CONSTEXPR uint32 dataAlignment = MALLOCMC_AP_SHRINK_DATAALIGNMENT;
+    static constexpr uint32 dataAlignment = MALLOCMC_AP_SHRINK_DATAALIGNMENT;
 
     // \TODO: The static_cast can be removed once the minimal dependencies of
     //        this project is are at least CUDA 7.0 and gcc 4.8.2
-    BOOST_STATIC_ASSERT(static_cast<uint32>(dataAlignment) > 0);
+    static_assert(dataAlignment > 0, "");
```
ax3l, line 77: I guess this will be the case now :)

src/include/mallocMC/alignmentPolicies/Shrink_impl.hpp
```diff
@@ -70,18 +68,18 @@ namespace Shrink2NS{
...
     //dataAlignment must also be a power of 2!
-    BOOST_STATIC_ASSERT(dataAlignment && !(dataAlignment & (dataAlignment-1)) ); 
+    static_assert(dataAlignment && !(dataAlignment & (dataAlignment-1)), "");
```
ax3l, line 79:
```suggestion
    static_assert(dataAlignment && !(dataAlignment & (dataAlignment-1)), "dataAlignment must also be a power of 2");
```

src/include/mallocMC/alignmentPolicies/Shrink_impl.hpp
ax3l, line 77: > at least CUDA 7.0 and gcc 4.8.2

I guess this will be the case now :)

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -623,6 +617,7 @@ namespace ScatterKernelDetail{
 #else
           unsigned int __mask = __ballot(1),
 #endif
+            // TODO(bgruber): why the leading __? those names are reserved
```
ax3l, line 620: The naming is a leftover from the original ScatterAlloc implementation. Should be no harm to name them properly, yep!

tests/verify_heap.cu
```diff
@@ -103,7 +103,7 @@ int main(int argc, char** argv){
 
   if( computeCapabilityMajor < 2 ) {
     std::cerr << "Error: Compute Capability >= 2.0 required. (is ";
-    std::cerr << computeCapabilityMajor << "."<< computeCapabilityMinor << ")" << std::endl;
+    std::cerr << computeCapabilityMajor << "."<< computeCapabilityMinor << ")\n";
```
sbastrakov, line 106: I think the change is good, to be consistent with the rest. Here it looks to not actually change anything as `cerr` always flushes.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: About the coding style: I guess nobody had time yet to put together a .clang-format file? That would enable automatic bulk formatting from the command line and some IDEs (VS for sure, CLion probably as well) can automatically pick up the file if it exists and format while typing.

sbastrakov: That's a great question! We had alpaka effort for it [here](https://github.com/alpaka-group/alpaka/pull/781). The issue is, we can't make it support template formatting like we want to have it, and a few of our projects heavily employ templates.

psychocoderHPC: @bernhardmgruber It would be nice to have clang format but we have heavily templated code in many project therefore the code style is a little bit special in the template formation part. Clang format has no useful rules to handle templates. For functions/methods you can place each argument on a seperate line and you can define how parentheses will be handled (east vs. west coast style). Template parentheses `<`, `>` will not be handled as parentheses :-(

psychocoderHPC:
I merged @ax3l's PR #170 which updated the CI. @bernhardmgruber you could rebase against the development branch.

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
```diff
@@ -76,19 +75,19 @@ namespace DistributionPolicies{
...
 #ifndef MALLOCMC_DP_XMALLOCSIMD_PAGESIZE
-#define MALLOCMC_DP_XMALLOCSIMD_PAGESIZE Properties::pagesize::value
+#define MALLOCMC_DP_XMALLOCSIMD_PAGESIZE (Properties::pagesize)
 #endif
-      BOOST_STATIC_CONSTEXPR uint32 pagesize      = MALLOCMC_DP_XMALLOCSIMD_PAGESIZE;
+      static constexpr uint32 pagesize      = MALLOCMC_DP_XMALLOCSIMD_PAGESIZE;
 
       //all the properties must be unsigned integers > 0
-      BOOST_STATIC_ASSERT(!std::numeric_limits<typename Properties::pagesize::type>::is_signed);
+      static_assert(!std::numeric_limits<decltype(Properties::pagesize)>::is_signed, ""); // TODO(bgruber): I think this is too strict. std::is_convertible_v should be enough
```
psychocoderHPC, line 83: What kind of advantages do we have if we relax it and use `std::is_convertible_v`

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -716,14 +711,15 @@ namespace ScatterKernelDetail{
       {
         uint32 linid = threadIdx.x + blockDim.x*(threadIdx.y + threadIdx.z*blockDim.y);
         uint32 threads = blockDim.x*blockDim.y*blockDim.z;
-        uint32 linblockid = blockIdx.x + gridDim.x*(blockIdx.y + blockIdx.z*gridDim.y);
+        const uint32 linblockid = blockIdx.x + gridDim.x*(blockIdx.y + blockIdx.z*gridDim.y);
         uint32 blocks =  gridDim.x*gridDim.y*gridDim.z;
-        linid = linid + linblockid*threads;
+        linid += linblockid*threads;
+
+        uint32 numregions = ((unsigned long long)memsize)/( ((unsigned long long)regionsize)*(sizeof(PTE)+pagesize)+sizeof(uint32)); // TODO(bgruber): why unsigned long long and not size_t?
```
psychocoderHPC, line 718: I think we want to be sure that this is 64bit and not overflow. `size_t` is on 32bit architectures 4 byte.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -744,7 +740,7 @@ namespace ScatterKernelDetail{
         uint32* regions = (uint32*)(ptes + numpages);
         //sec check for mem size
         //this check refers to the original memory-pointer, which was not adjusted!
-        if( (void*)(regions + numregions) > (((char*)memory) + memsize) )
+        if( (void*)(regions + numregions) > (((char*)memory) + memsize) ) // TODO(bgruber): cast to void* is confusing and looks dangerous
```
psychocoderHPC, line 743: Feel free to cast the left hand side to `char*` too

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
```diff
@@ -76,19 +75,19 @@ namespace DistributionPolicies{
...
 
       // \TODO: The static_cast can be removed once the minimal dependencies of
```
psychocoderHPC, line 85: please remove this comment

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
```diff
@@ -76,19 +75,19 @@ namespace DistributionPolicies{
...
       //        this project is are at least CUDA 7.0 and gcc 4.8.2
-      BOOST_STATIC_ASSERT(static_cast<uint32>(pagesize) > 0);
+      static_assert(pagesize > 0, "");
```
psychocoderHPC, line 87: Could you please add `page size of zero is not allowed`.

src/include/mallocMC/mallocMC_constraints.hpp
```diff
@@ -80,8 +79,7 @@ namespace mallocMC{
     typename CreationPolicies::Scatter<x,y>,
     typename DistributionPolicies::XMallocSIMD<z> 
   >{
-    BOOST_MPL_ASSERT_MSG(x::pagesize::value == z::pagesize::value,
-        Pagesize_must_be_the_same_when_combining_Scatter_and_XMallocSIMD, () );
+    static_assert(x::pagesize == z::pagesize, "Pagesize_must_be_the_same_when_combining_Scatter_and_XMallocSIMD");
```
psychocoderHPC, line 82: could you please remove the underscores. `_`

psychocoderHPC:
Could you please change also that we require CUDA 9+ else C++14 is not supported.

https://github.com/ComputationalRadiationPhysics/mallocMC/blob/e383f3cd892420839bc9ea3dd1f2d41a2505332f/CMakeLists.txt#L24-L31

should be
```
find_package(CUDA "9.0" REQUIRED)
if(NOT DEFINED COMPUTE_CAPABILITY)
        set(COMPUTE_CAPABILITY "30")
endif()
```

tests/verify_heap.cu
bernhardmgruber: You are right!

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: done

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
bernhardmgruber: removed

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
bernhardmgruber: No, because I just removed the static_assert as by your previous comment ;)
To further elaborate on not having a message: I am very used to C++17, where the message is optional and I usually do not provide one if the checked condition is simple. Furthermore, there are also no messages attached to runtime assert(...)s in the code.

src/include/mallocMC/mallocMC_constraints.hpp
bernhardmgruber: makes perfect sense!

bernhardmgruber:
> Could you please change also that we require CUDA 9+ else C++14 is not supported.
> 
> https://github.com/ComputationalRadiationPhysics/mallocMC/blob/e383f3cd892420839bc9ea3dd1f2d41a2505332f/CMakeLists.txt#L24-L31
> 
> should be
> 
> ```
> find_package(CUDA "9.0" REQUIRED)
> if(NOT DEFINED COMPUTE_CAPABILITY)
>         set(COMPUTE_CAPABILITY "30")
> endif()
> ```

I did that, but I omitted the quotes around the CUDA version. Seems to work fine and I found a sample on the web doing it the same way.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: done

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: I renamed PAGE to Page.

Hmm, what was the last version of clang-format that you have tried? There are constant improvements.
And thank you for the styleguide instructions! But I work in too many projects (also hobby projects), which all have different coding styles. So clang-format files in every repository work superbly here! So having one would already help a lot, even if it's not perfect! I will have a look at the file for alpaka.

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
bernhardmgruber: The advantage is that 4096 is a valid page size, not only 4096u. See e.g. the following configuration:
```
struct ScatterConfig{
  static constexpr auto pagesize = 4096u;
  static constexpr auto accessblocks = 8;
  static constexpr auto regionsize = 16;
  static constexpr auto wastefactor = 2;
  static constexpr auto resetfreedpages = false;
};
```
Only pagesize needs to be unsiged in order to compile. I think this is a usability bug.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: Hmm, firstly: Do you encounter 32bit system because CUDA is a different architecture than the host and could be 32bits? 

Secondly, assuming the RHS of the divison could be larger than the domain of size_t and thus be larger than the maximum value of the LHS, which is of type size_t, the result would be 0. Thus numregions, and numpages would be 0. And then I guess the whole allocator would not work?
So if the RHS can never assume a value greater than the maximum size_t can hold without loosing its meaning, the computation can be done in size_t.

Since the RHS is the size of a memory region, one could also argue that we should assert that it is not larger than the entire allocated pool (ie. memsize).

What do you think?

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
ax3l: Shouldn't we make them all unsigned instead? :D

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
ax3l: Upsi.

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
psychocoderHPC: I am fine with changing it to signed but we should take care that we avoid as good as possible implicit casts, even if the code is currently not checked to avoid it.

bernhardmgruber: I still don't think that this is a problem, since the class member to which the signed value is assigned to, is still unsigned (see line 80):
```
static constexpr uint32 pagesize      = MALLOCMC_DP_XMALLOCSIMD_PAGESIZE;
```

bernhardmgruber: @ax3l: in principal, all the values are semantically unsigned. But inside the policy types I just want to carry literals. So any data type is fine for me, as long as it convertes to what I expect in the implementation later. Otherwise it is a configuration error, which the compiler tells you ;)
And the implementation seems to use uint32_t internally.
About signed ints: they benefit from UB wrt. wrapping behavior so the optimizer can make more assumptions, which can help.

bernhardmgruber:
I modernized the CMakeLists.txt a bit and now the CI build succeeds. The main problem was that C++14 was not enabled for nvcc.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
psychocoderHPC: We used PIConGPU with mallocMC few years ago on a jetson (32bit). Out of the box I can not say if 32bit is still required. I would keep it open and shift this designation to a later PR.

CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
-project(mallocMC)
-cmake_minimum_required(VERSION 2.8.12.2)
+project(mallocMC LANGUAGES CUDA CXX)
+cmake_minimum_required(VERSION 3.8)
+
+set(CMAKE_CXX_STANDARD 14)
```
psychocoderHPC, line 4: Do we use a feature which requires C++14 if not we should require C++11

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: I see. I will remove the TODO then and let's keep the code as is.

CMakeLists.txt
bernhardmgruber: done

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: After a talk with @psychocoderHPC today I will try to get a .clang-format file working at some point. But I think it does not need to be part of this PR.

psychocoderHPC:
> Wow, this is so awesome, thanks a lot!   And you refactored out Boost completely, oh yeah!
> 
> Feel free to bump the CI to require CUDA 9.2+. Anything else is considered ancient and unusable for proper C++14.

Note: Offline discussed with @bernhardmgruber, we will require C++11 only since there is no  C++14 feature in the code base.

This allow us to create a release with C++11 and switch to C++14 with the alpaka refactored version of mallocMC.

PR #171, opened 2020-05-11
Adding headers to projects and applied clang-tidy

bernhardmgruber:
* added a custom target for mallocMC headers, this makes the headers show up in IDE projects (e.g. Visual Studio)
* added header files to their respective projects (exampels and VerifyHeap)
* applied clang-tidy
* using global includes in examples and tests for mallocMC headers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,6 @@ if(NOT DEFINED COMPUTE_CAPABILITY)
   set(COMPUTE_CAPABILITY "30")
 endif()
 set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -arch=sm_${COMPUTE_CAPABILITY} -use_fast_math")
-include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 OPTION(CUDA_OUTPUT_INTERMEDIATE_CODE "Output ptx code" OFF)
 if(CUDA_OUTPUT_INTERMEDIATE_CODE)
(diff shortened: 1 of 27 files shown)


CMakeLists.txt
```diff
@@ -72,9 +71,13 @@ INSTALL(
...
 # Executables
 ###############################################################################
-add_custom_target(examples DEPENDS mallocMC_Example01 mallocMC_Example02 mallocMC_Example03 VerifyHeap)
+file(GLOB_RECURSE headers src/include/**)
+add_custom_target(mallocMC SOURCES ${headers}) # create a target with the header files for IDE projects
+source_group(TREE ${CMAKE_CURRENT_LIST_DIR}/src/include FILES ${headers})
 
-add_executable(mallocMC_Example01 EXCLUDE_FROM_ALL examples/mallocMC_example01.cu)
+include_directories(${CMAKE_CURRENT_LIST_DIR}/src/include)
+add_executable(mallocMC_Example01 EXCLUDE_FROM_ALL examples/mallocMC_example01.cu examples/mallocMC_example01_config.hpp)
```
bernhardmgruber, line 79: btw, why do you specify the EXCLUDE_FROM_ALL? I know it's a header-only library. So the most users don't need to build anything. So we can just as well build the examples and tests if someone builds the project (either running `make` or hitting build all in an IDE).

psychocoderHPC: The CMake is very old. This is the workaround taht the test will not be build when we use mallocMC in a third party project. 
In the past we thought there is no need to install a header only library. This thinking is wrong. We need to refactore the CMake file to allow installing the header only library via  `make install` or the cmake equivalent.

bernhardmgruber: I see. Thank you! I think we should properly rework the CMakeLists at some point then. But probably not as part of this PR.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -75,7 +75,7 @@ namespace ScatterKernelDetail{
   {
     public:
       typedef T_Config  HeapProperties;
```
psychocoderHPC, line 77: Somehow clang-tidy missed to change this. Could you maybe grep for `typedef` and adjust this and other lines by hand.

bernhardmgruber: thx, done!

PR #172, opened 2020-05-14
clang-format

bernhardmgruber:
* added a .clang-format file, based on the one suggested for alpaka: https://github.com/alpaka-group/alpaka/pull/781
* formatted whole codebase
* enabled trailing return types in clang-tidy and applied fix

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,76 @@
+---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: AlwaysBreak
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: DontAlign
+AlignOperands: false
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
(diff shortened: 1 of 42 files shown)


psychocoderHPC:
We need to document in the readme or contribution the command you need to use to format all files.

bernhardmgruber:
Shouldn't that be obvious by the existance of the .clang-format file?

Otherwise, what would you prefer? Creating a new contributing.md, add it to readme.md, where there is no developer related content, or install.md where the build instructions are?

As for the command, that depends a lot on what shell and operating system you use. I think a developer should find this information for themselves on the web.

psychocoderHPC:
> Shouldn't that be obvious by the existence of the .clang-format file?

No it is not because the command is not simply `clang-format`, do not assume that someone knows a special tool.

> Otherwise, what would you prefer? Creating a new contributing.md, add it to readme.md, where there is no developer related content, or install.md where the build instructions are?

I see we do not have such a file, so I suggest to create it and link it in `README.md` too.

> As for the command, that depends a lot on what shell and operating system you use.

You can provide it for linux and windows. MAC should be equal to windows.

> I think a developer should find this information for themselves on the web.

You can provide a link to the documentation but give them a copy past line too.

bernhardmgruber:
I added a CONTRIBUTING.md and described briefly how to format.

psychocoderHPC:
Note: We will take this code style for now. It could be we need to update this file again after we agreed for a common style file in https://github.com/ComputationalRadiationPhysics/contributing/issues/33

Issue #166, opened 2020-02-06
HIP support


psychocoderHPC:
I am currently working on a native version for HIP https://github.com/psychocoderHPC/picongpu/tree/topic-hip-amd-benachmark/thirdParty/mallocMC.

This is only a prototype, I will port it soon to CUPLA to support different backend e.g. CUDA/HIP and CPU.

psychocoderHPC:
for the record: @saipavankalyan also started to port mallocMC and reported linker issues with HIP-hcc https://github.com/ROCm-Developer-Tools/HIP/issues/1849

psychocoderHPC:
HIP prototype https://github.com/psychocoderHPC/mallocMC/tree/topic-hip-port
Currently the cmake is not updated

psychocoderHPC:
- HIP-nvcc
```
cmake ../mallocMC/ -D mallocMC_HIP_PLATFORM=nvcc
make -j examples
```
 https://github.com/psychocoderHPC/mallocMC/tree/topic-hip-port contains now support for all variations of HIP.  ATTENTION: HIP-hcc uses a workaround where the mallocMC heap is not freed when the heap object is destructed. I added the support only as proof of work, you should not use HIP-hcc anymore.

- HIP-clang
```
cmake ../mallocMC/ -D mallocMC_HIP_PLATFORM=clang
make -j examples
```

- HIP-hcc
```
cmake ../mallocMC/ -D mallocMC_HIP_PLATFORM=hcc
make -j examples
```

chillenzer:
We can close this, right?

PR #173, opened 2020-05-14
Port to alpaka

bernhardmgruber:
* integrated some changes from hipifycation in alpaka
* replaced mallocMC CUDA macros by alpaka macros, removed mallocMC_prefixes.hpp
* replaced all CUDA kernel invocations by alpaka kernel enqueues
* removed all code that targetted CUDA < 9
* merged example02 into example01 since they are almost the same
* inlined content of mallocMC_example01_config.hpp
* ported kernel invocations to alpaka
* replaced cuda allocation routines by alpaka
* renamed .cu source files to .cpp
* reworked CMakeLists.txt (removed all CUDA stuff, removed big block comments, ...)
* added new ReservePoolPolicies SimpleMalloc, intended for running allocator in host memory
* passing Alpaka Accelerator through almost all device functions
* replaced all atomit operations by alpaka atomics
* replaced all CUDA intrinsics by custom implementations in mallocMC_utils.hpp, which default to the intrinsics of the corresponding platform or a default CPU implementation
* tried to #ifdef some CUDA thread sync primitives
* replaced CUDA thread IDs with alpaka indices and workdivs
* replaced __shared__ memory by alpaka shared allocVar
* SimpleCudaMalloc and XMallocSIMD are not available, when CUDA is not available, because they are too hard to port for now
* refactored thread indexing
* incorporating changes from psychocoderHPC from: https://github.com/ComputationalRadiationPhysics/mallocMC/compare/dev...psychocoderHPC:topic-hip-port
* added a target mallocMCIde to CMakeLists.txt, so developers can browse the code in IDEs
* setting compiler warnings via a warnings target, instead of global CMAKE_CXX_FLAGS
* setting include directories on targets instead of globally
* removed check for CUDA compute capability, since capability 3 is required since CUDA 9
* removed cudaSetDeviceFlags, as it's not needed

- [ ] remove workaround commit cd97fe83c8f301024c535d746db47d80025a273a (return type for alpaka min/max)


CMakeLists.txt
```diff
@@ -1,83 +1,50 @@
...
-
-SET(CUDA_OPTIMIZATION_TYPE "unset" CACHE STRING "CUDA Optimization")
-set_property(CACHE CUDA_OPTIMIZATION_TYPE PROPERTY STRINGS "unset;-G0;-O0;-O1;-O2;-O3")
-if(NOT ${CUDA_OPTIMIZATION_TYPE} STREQUAL "unset")
-  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_OPTIMIZATION_TYPE}")
+    cmake_policy(SET CMP0074 NEW)
 endif()
 
+# find alpaka
+set(ALPAKA_ROOT "${CMAKE_CURRENT_LIST_DIR}/alpaka/lib/cmake/alpaka" CACHE STRING "The location of alpakaConfig.cmake")
```
psychocoderHPC, line 12: Please have a look to https://github.com/alpaka-group/cupla/blob/b72833cbf0844664b26bbf553289356cfa03c18a/cuplaConfig.cmake#L112-L127
This is the way how we integrate alpaka in cupla. It gives you also easily a way to switch away from the internal shipped alpaka implementation.

psychocoderHPC: `add_subdirectory` is the way to use a internal version. An external will be used via `find_package`.

CMakeLists.txt
```diff
@@ -1,83 +1,50 @@
...
+list(APPEND CMAKE_PREFIX_PATH "${ALPAKA_ROOT}")
+find_package(alpaka REQUIRED)
 
-################################################################################
-# Warnings
-################################################################################
+# warnings
 if(CMAKE_COMPILER_IS_GNUCXX)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wshadow -Wno-unknown-pragmas -Wextra -Wno-unused-parameter -Wno-unused-local-typedefs")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wshadow -Wno-unknown-pragmas -Wextra -Wno-unused-parameter -Wno-unused-local-typedefs")
```
psychocoderHPC, line 18: Note: Nothing we should change in this PR. Those flags should only be added when the CI is running. These flags are global and will effect the user code too even if the user is not interested to fulfilling it.

ax3l:
Just curious because there is no issue open describing this: why do we port mallocMC to Alpaka?

psychocoderHPC:
The main motivation is to integrate it into PIConGPU. I have a prototype for HIP but do not like to port mallocMC to any possible platform.

psychocoderHPC:
For the documentation: there is an open issue to port mallocMC to HIP #166

CMakeLists.txt
bernhardmgruber: Thank you! I added alpaka develop as a git subtree and took the cmake code from cupla.

CMakeLists.txt
bernhardmgruber: I created a target "warnings" and added the flags there. The target is linked in all examples and the VerifyHeap target.

bernhardmgruber:
@psychocoderHPC the CI build now fails because the cmake version on the CI slave is too old. alpaka requires cmake 3.15 or newer. can you have a look at this for me please? thank you!

psychocoderHPC:
> @psychocoderHPC the CI build now fails because the cmake version on the CI slave is too old. alpaka requires cmake 3.15 or newer. can you have a look at this for me please? thank you!

NP, I will update the travis script tomorrow.

src/include/mallocMC/distributionPolicies/XMallocSIMD_impl.hpp
```diff
@@ -33,10 +33,12 @@
 
 #pragma once
 
-#include "../mallocMC_prefixes.hpp"
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
```
bernhardmgruber, line 36: I had a hard time getting XMallocSIMD working on CPU platforms, so I eventually just disabled it. Is there even a point of having it available on CPUs?

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
```diff
@@ -27,6 +27,8 @@
 
 #pragma once
 
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
```
bernhardmgruber, line 30: I also disabled this for non-CUDA targets. Is there any way I can implement this for CPU targets?

src/include/mallocMC/reservePoolPolicies/SimpleCudaMalloc_impl.hpp
```diff
@@ -27,6 +27,8 @@
 
 #pragma once
 
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
```
bernhardmgruber, line 30: Can I implement this for CPU targets? E.g. by just using ::malloc()? For now I created a separate reservePolicy, but maybe we should just rename the whole policy into DeviceMalloc and use alpaka buffers. What do you think?

src/include/mallocMC/reservePoolPolicies/SimpleMalloc.hpp
```diff
@@ -1,11 +1,5 @@
 /*
   mallocMC: Memory Allocator for Many Core Architectures.
-  https://www.hzdr.de/crp
-
-  Copyright (C) 2014 Institute of Radiation Physics,
-                     Helmholtz-Zentrum Dresden - Rossendorf
-
-  Author(s):  Carlchristian Eckert - c.eckert ( at ) hzdr.de
```
bernhardmgruber, line 8: I guess you want something here from me? Like copyright CASUS and CERN and my name.

tests/verify_heap.cpp
```diff
@@ -107,34 +92,32 @@ static constexpr size_t heapInMB_default = 1024; // 1GB
...
         &computeCapabilityMinor, cudaDevAttrComputeCapabilityMinor, 0);
-
     if(computeCapabilityMajor < 2)
     {
         std::cerr << "Error: Compute Capability >= 2.0 required. (is ";
         std::cerr << computeCapabilityMajor << "." << computeCapabilityMinor
                   << ")\n";
         return 1;
     }
+#endif
```
bernhardmgruber, line 117: Any way alpaka can check this? Otherwise: do we even need it?

tests/verify_heap.cpp
```diff
@@ -426,10 +430,13 @@ __global__ void deallocAll(
  *
  * @param data the datastructure to damage
  */
-__global__ void damageElement(allocElem_t ** data)
+struct DamageElement
```
bernhardmgruber, line 433: By any chance, can I just define the kernel as a normal function template? I could not get it compiling. Only a struct or a lambda seem to work with alpaka.

tests/verify_heap.cpp
```diff
@@ -664,7 +660,11 @@ auto run_heap_verification(
     const unsigned threads,
     const bool machine_readable) -> bool
 {
-    cudaSetDeviceFlags(cudaDeviceMapHost);
```
bernhardmgruber, line 667: Is that still needed? I removed it because it seemed to not make a difference.

tests/verify_heap.cpp
```diff
@@ -664,7 +660,11 @@ auto run_heap_verification(
...
+    const auto dev = alpaka::pltf::getDevByIdx<alpaka::pltf::Pltf<Device>>(0);
+    auto queue = Queue{dev};
+
+    // cudaSetDeviceFlags(cudaDeviceMapHost); // TODO alpaka, and do we actually
+    // need this?
```
bernhardmgruber, line 667: Do we need it?

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -290,21 +259,25 @@ namespace mallocMC
...
             {
                 // get first spot
                 uint32 spot = randInit() % spots;
                 for(;;)
                 {
                     const uint32 mask = 1 << spot;
-                    const uint32 old = atomicOr(bitfield, mask);
+                    const uint32 old
+                        = alpaka::atomic::atomicOp<alpaka::atomic::op::Or>(
+                            acc, bitfield, mask);
```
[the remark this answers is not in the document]
sbastrakov: Sure! We generally tend to discuss such things as issues.

src/include/mallocMC/allocator.hpp
```diff
@@ -111,90 +130,120 @@ namespace mallocMC
...
          */
-        MAMC_HOST
-        void free()
+        template<typename AlpakaDevice>
+        ALPAKA_FN_HOST void free(AlpakaDevice & dev)
         {
-            cudaFree(allocatorHandle.devAllocator);
+            using Idx = alpaka::idx::traits::IdxType<AlpakaAcc>::type;
+            devAllocatorStorage = alpaka::mem::buf::alloc<DevAllocator, Idx>(
+                dev, Idx{0}); // TODO(bgruber): how to free?
```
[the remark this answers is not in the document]
sbastrakov: Yes, free happens at destructor.

tests/verify_heap.cpp
sbastrakov: Alpaka always gets a kernel is a functor. You can still have your implementation in a free function style, like in CUDA. Then just wrap it into a functor (lambda or `std::function`) when you create a kernel object instance to give to alpaka.

src/include/mallocMC/allocator.hpp
```diff
@@ -111,90 +130,120 @@ namespace mallocMC
...
     private:
-        AllocatorHandle allocatorHandle;
+        std::function<void()>
+            freeLastAlloc; // calles free(dev), with the dev used during alloc
+        decltype(alpaka::mem::buf::alloc<
+                 DevAllocator,
+                 alpaka::idx::traits::IdxType<AlpakaAcc>::type>(
+            std::declval<const alpaka::dev::Dev<AlpakaAcc>>(),
+            std::declval<
+                const alpaka::idx::traits::IdxType<AlpakaAcc>::type>()))
```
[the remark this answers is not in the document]
sbastrakov: Yes, these long types are an unfortunate side of alpaka. We often try to manage it by first creating a helper `using` type aliases with readable name, to be used later in more complex constructions. Something along the lines of
```cpp
using Dim = alpaka::dim::Dim<AlpakaAcc>;
using Idx = alpaka::idx::Idx<AlpakaAcc>;
using Dev = alpaka::dev::Dev<AlpakaAcc>;
using Buf = alpaka::mem::buf::Buf<Dev, YourDataType, Dim, Idx>;
```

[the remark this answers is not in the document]
psychocoderHPC: > Because that decltype is just too big.

What do you mean by `too big`?

> what is the type of a buffer in alpaka?

Do you mean how to create a buffer with alpaka? https://github.com/alpaka-group/alpaka/blob/a5a8277cded3e20612c11282857ad2f007a113f3/example/vectorAdd/src/vectorAdd.cpp#L137-L138

src/include/mallocMC/allocator.hpp
```diff
@@ -111,90 +130,120 @@ namespace mallocMC
...
+            devAllocatorStorage; // TODO(bgruber): since we want to have an
+                                 // empty/invalid buffer sometimes, this should
+                                 // be std::optional<>
```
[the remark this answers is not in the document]
psychocoderHPC: Create an buffer with the size zero. We removed the dependency to boost so I would avoid to use it even if alpaka depend on it. `std::optional` is C++17 :-(
So I suggest a pointer :-( or the zero sized buffer.

src/include/mallocMC/allocator.hpp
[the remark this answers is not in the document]
bernhardmgruber: @psychocoderHPC: I mean this declaration I have put into Allocator:
```
decltype(alpaka::mem::buf::alloc<
                 DevAllocator,
                 alpaka::idx::traits::IdxType<AlpakaAcc>::type>(
            std::declval<const alpaka::dev::Dev<AlpakaAcc>>(),
            std::declval<
                const alpaka::idx::traits::IdxType<AlpakaAcc>::type>()))
            devAllocatorStorage;
```

@sbastrakov: I will try your using declaration!

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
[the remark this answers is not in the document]
bernhardmgruber: https://github.com/alpaka-group/alpaka/issues/1005

src/include/mallocMC/allocator.hpp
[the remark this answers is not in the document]
sbastrakov: Yes, 7 line `decltype` qualifies as too big, even for alpaka standards xd.

src/include/mallocMC/reservePoolPolicies/SimpleMalloc.hpp
psychocoderHPC: yes would be nice

tests/verify_heap.cpp
bernhardmgruber: Ok, thx for the explanation!

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -577,39 +580,53 @@ namespace mallocMC
...
+                                pagesize);
                         if(old == 0)
                         {
                             // clean the bits for the hierarchy
                             _page[page].init();
                             // remove chunk information
                             _ptes[page].chunksize = 0;
-                            __threadfence();
+#if defined(__HIP__) || defined(_CUDA_ARCH_)
+                            __threadfence(); // TODO alpaka
```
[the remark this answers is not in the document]
sbastrakov: Yes, look at `alpaka/core/BoostPredef.hpp`

tests/verify_heap.cpp
psychocoderHPC: As I know alpaka is not providing this information. Note that ` if(computeCapabilityMajor < 2)` is not required since CUDA 9.0 compute architecture is at least 3.0

tests/verify_heap.cpp
psychocoderHPC: `std::function` should not work on device.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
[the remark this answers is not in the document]
sbastrakov: Oh wait, i guess you meant` __threadfence()`, not the ifdef. For the former I am not sure if alpaka has a counterpart.

tests/verify_heap.cpp
psychocoderHPC: Yes remove it  I do not think it is required

tests/verify_heap.cpp
psychocoderHPC: I do not think so.

psychocoderHPC:
@bernhardmgruber  Could you restructure your commits. So that the alpaka subtree creation is the first commit and all your mallocMC changes are in a second. Currently it is not possible to review  review your changes.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -939,8 +1003,24 @@ namespace mallocMC
...
+                                      size_t memsize) {
+                    heap->pool = heapmem;
+                    heap->initDeviceFunction(acc, heapmem, memsize);
+                };
+                // TODO(bgruber): how do I know which dimension the user's
+                // AlpakaAcc is using?
+                using Dim = alpaka::dim::traits::DimType<AlpakaAcc>::type;
+                using Idx = alpaka::idx::traits::IdxType<AlpakaAcc>::type;
+                const auto workDiv = alpaka::workdiv::WorkDivMembers<Dim, Idx>{
+                    Idx{1}, Idx{256}, Idx{1}};
```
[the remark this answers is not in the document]
sbastrakov: I do not think it works. I believe the way is to provide three values of type `alpaka::vec::Vec<Dim, Idx>` to the `workDiv ` constructor. And such a vector can't be constructed from one value in 2D and 3D. A possible, but clunky way to do that is to get a vector of 1 using its static method `ones()` and then set one of its elements to what you want.

src/include/mallocMC/reservePoolPolicies/SimpleMalloc.hpp
bernhardmgruber: done

src/include/mallocMC/allocator.hpp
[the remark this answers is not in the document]
bernhardmgruber: improved

tests/verify_heap.cpp
bernhardmgruber: Thx! I removed the check.

tests/verify_heap.cpp
bernhardmgruber: Great!

tests/verify_heap.cpp
bernhardmgruber: Removed

bernhardmgruber:
@psychocoderHPC I just tried, but it seems I can no longer rebase with the git subtree :/ whenever git rebase processes the addition of the alpaka subtree, it tries to integrate all changes into the root working copy instead of the alpaka subfolder. and that messes up everything. I will try to come up with something different.

src/include/mallocMC/allocator.hpp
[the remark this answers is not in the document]
bernhardmgruber: I went with a std::unique_ptr for now, but I need to say that this hurts my heart! optional is incredibly useful and working without it in mid 2020 is a crime.

src/include/mallocMC/allocator.hpp
```diff
@@ -111,90 +130,120 @@ namespace mallocMC
...
             DistributionPolicy,
             OOMPolicy,
             AlignmentPolicy>;
-        using AllocatorHandle = AllocatorHandleImpl<Allocator>;
+        using AllocatorHandle = AllocatorHandleImpl<AlpakaAcc, Allocator>;
 
     private:
-        AllocatorHandle allocatorHandle;
+        std::function<void()>
+            freeLastAlloc; // calles free(dev), with the dev used during alloc
```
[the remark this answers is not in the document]
bernhardmgruber: dropped, because `devAllocatorStorage` is now an "optional".

psychocoderHPC:
> @psychocoderHPC I just tried, but it seems I can no longer rebase with the git subtree :/ whenever git rebase processes the addition of the alpaka subtree, it tries to integrate all changes into the root working copy instead of the alpaka subfolder. and that messes up everything. I will try to come up with something different.

Yes I know that is always the case with subtrees.
The workflow is
- remove the commit where you introduced alpaka
- stash all other commits
- init/update the alpaka subtree
- stash pop your changes


@bernhardmgruber Could you please squash all your commits into one commit. Currently it is still hard to review the changes.

bernhardmgruber:
> The workflow is
> 
> * remove the commit where you introduced alpaka
> * stash all other commits
> * init/update the alpaka subtree
> * stash pop your changes

Thx! I created a new branch alpaka2 from dev and added the alpaka subtree there. Then I rebased the subtree out of the alpaka branch. And then I put alpaka on top of the alpaka2. But your solution is probably easier.

> @bernhardmgruber Could you please squash all your commits into one commit. Currently it is still hard to review the changes.

I can do that. But can you give me a bit of a rational? This looses all the intermediate changes I did. Or are they of no interest? The resulting diff will also be huge.

psychocoderHPC:
> I can do that. But can you give me a bit of a rational? This looses all the intermediate changes I did. Or are they of no interest? The resulting diff will also be huge.

Currently  if I try to review the PR I need to got commit by commit to 22 commits. I will see changes you have already reverted in newer commits. If it end with 3 commits to review it's also ok if you group it by any logic. Since we introduced alpaka here and we can not deselect this commit and review the full diff it is the only way.
An other option would be to push the alpaka subtree in  a separate PR and switch in the same moment to C++14.

bernhardmgruber:
@psychocoderHPC: I can see that the diff is unwieldy with the addition of the alpaka subtree. So this I guess we should split this out anyway. Here is a separate PR: https://github.com/ComputationalRadiationPhysics/mallocMC/pull/176

And for reviewing the diff, don't you use the Github Files changed tag? It provides a nice unified diff.

bernhardmgruber:
@psychocoderHPC I spashed all changes.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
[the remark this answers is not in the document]
psychocoderHPC: no we have currently no function for that.

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -111,14 +59,26 @@ namespace mallocMC
         using type = unsigned long long;
     };
 
+#if defined(__HIP__) || defined(_CUDA_ARCH_)
```
psychocoderHPC, line 62: it must be `__CUDA_ARCH__` everywhere.

psychocoderHPC: and `__HIP_DEVICE_COMPILE__== 1 && defined(__HIP__)` to check if you are on a HIP device.

https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary

`__HIP_DEVICE_COMPILE__` is undefined on the host side

src/include/mallocMC/reservePoolPolicies/SimpleCudaMalloc_impl.hpp
psychocoderHPC: Yes o CPU it can become a simple `::malloc` 

Maybe we should call this policy `SimpleMalloc` or `SystemMalloc` and use the cuda/hip way or ::malloc.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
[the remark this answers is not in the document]
bernhardmgruber: It works btw! If I specify a workDiv with a lower dimension, the other dimensions are set to 1, which seems reasonable to me. What is less intuitive now is how you get the index and worksize in the kernel:
```
const auto id = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc).sum();
const auto n = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc).prod();
```
This should work, because the index contains 0 for all unused dimensions, and the workDiv contains 1 for all unused dimensions.
I would have liked to just use `[0]` from index and workDiv, but unfortunately alpaka indices are stored `[x]`, `[y, x]` or `[z, y, x]` and checking the dimension and indexing is way more code than `sum()` and `prod()`.

src/include/mallocMC/mallocMC_utils.hpp
bernhardmgruber: your link suggests that just checking if `__HIP_DEVICE_COMPILE__` is defined is enough.

tests/dimensions.cpp
```diff
@@ -0,0 +1,383 @@
...
+    test2D<alpaka::acc::AccGpuCudaRt>();
+    test3D<alpaka::acc::AccGpuCudaRt>();
+
+    test1D<alpaka::acc::AccCpuThreads>();
+    test2D<alpaka::acc::AccCpuThreads>();
+    test3D<alpaka::acc::AccCpuThreads>();
+
+    test1D<alpaka::acc::AccCpuOmp2Threads>();
+    test2D<alpaka::acc::AccCpuOmp2Threads>();
+    test3D<alpaka::acc::AccCpuOmp2Threads>();
```
bernhardmgruber, line 376: I would like to have a unit test framework for running these. I saw alpaka uses Catch2. Would you and how would you integrate it here as well?
If you are not certain about Catch2, I am also a big fan of gtest.

src/include/mallocMC/reservePoolPolicies/Malloc.hpp
```diff
@@ -0,0 +1,55 @@
...
+        {};
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+        template<typename Dim, typename Idx>
+        struct Malloc<alpaka::acc::AccGpuCudaRt<Dim, Idx>> : SimpleCudaMalloc
+        {};
+#endif
+#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+        template<typename Dim, typename Idx>
+        struct Malloc<alpaka::acc::AccGpuHipRt<Dim, Idx>> : SimpleCudaMalloc
+        {};
```
bernhardmgruber, line 52: I am in strong favor of merging SimpleMalloc and SimpleCudaMalloc into Malloc.

src/include/mallocMC/reservePoolPolicies/SimpleCudaMalloc_impl.hpp
bernhardmgruber: I called it Malloc.

alpaka/include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
```diff
@@ -71,6 +71,9 @@ namespace alpaka
                     MaxUniformCudaHipBuiltIn const & max_ctx,
                     Tx const & x,
                     Ty const & y)
+#ifdef _MSC_VER
+                    -> Tx // FIXME(bgruber): return type is deduced as void by MSVC as host compiler (nvcc deduces correct return type)
+#endif
```
bernhardmgruber, line 76: I have no clue how to solve this. Maybe it is an issue with MSVC. Maybe also in the CUDA SDK for windows. Any ideas?

bernhardmgruber: Funnily enough, I just tried the whole thing on Gentoo with CUDA 10.2 and have the same issue with g++8.4 as host compiler:
`error: void value not ignored as it ought to be`

tests/dimensions.cpp
sbastrakov: Yes, alpaka uses Catch2 and I think the idea was that other projects that did not commit to any framework yet should (but not have to) also prefer this. What do you think @psychocoderHPC ?

alpaka/include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
sbastrakov: I do not have an idea, but can try on my local machine with MSVS + CUDA. It may be that something does wrong with `::max(x, y)` that MSVS and g++ 8.4 for some reason think it returns `void`?

tests/dimensions.cpp
bernhardmgruber: I adopted Catch2 and split main() into several test cases.

src/include/mallocMC/reservePoolPolicies/Malloc.hpp
bernhardmgruber: Actually, we should go the alpaka way and just drop the reservePoolPolicy entirely. The buffer is then allocated for the device passed into the ctor of Allocator.

bernhardmgruber: Dropped all reserve pool policies and replaced them by an alpaka buffer.

alpaka/include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
bernhardmgruber: I eventually just explicitely specified the return type as `decltype(::max(x, y))`. It seems this fixes the misscompilation.

bernhardmgruber:
I added the tests to the CI as well and now the Catch main fails to compile. Likely because it is preprocessed by nvcc. Any ideas?
Alpaka seems to have a separate library for the test main. Can I disable alpaka_add_executable from treating all cpp files as CUDA files?

alpaka/include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
psychocoderHPC: Is this change only a preview?
Please do not change any line in a subtree, this change will get lost after with the next subtree update.

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -128,74 +88,64 @@ namespace mallocMC
...
      */
-    MAMC_ACCELERATOR inline auto warpid() -> std::uint32_t
-    {
+    ALPAKA_FN_ACC inline auto warpid()
+    {
+#if defined(__HIP_DEVICE_COMPILE__) && defined(__HIP__)
+        // get wave id
+        // https://github.com/ROCm-Developer-Tools/HIP/blob/f72a669487dd352e45321c4b3038f8fe2365c236/include/hip/hcc_detail/device_functions.h#L974-L1024
+        return __builtin_amdgcn_s_getreg(GETREG_IMMED(3, 0, 4));
+#elif defined(__HIP_DEVICE_COMPILE__) && defined(__HCC__)
```
psychocoderHPC, line 97: You can remove all `HCC` branches, alpaka is only supporting `HIP-clang` and `HIP-nvcc`
Could you please write `#if defined(__CUDA_ARCH__)` first and HIP as `#elif defined(__HIP_DEVICE_COMPILE__) && defined(__HIP__)`, the result will be that `HIP-nvcc` code performs equal to native CUDA code.

Please apply this change to all utility functions.

examples/mallocMC_example03.cpp
```diff
@@ -65,49 +60,60 @@ struct ScatterHashParams
...
-// configure the AlignmentPolicy "Shrink"
 struct AlignmentConfig
 {
     static constexpr auto dataAlignment = 16;
 };
 
-// Define a new mMCator and call it ScatterAllocator
-// which resembles the behaviour of ScatterAlloc
 using ScatterAllocator = mallocMC::Allocator<
+    Acc,
```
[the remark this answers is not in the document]
bernhardmgruber: I need the alpaka Acc for:
* the type of the buffers inside mallocMC::Accelerator. I could typ erase them maybe.
* the call to `alpaka::kernel::createTaskKernel` inside `initHeap` in the creation policy. If I can get the accelerator type from the alpaka device or queue, I could solve this otherwise. Is this possible?

src/include/mallocMC/mallocMC_utils.hpp
bernhardmgruber: done

examples/mallocMC_example03.cpp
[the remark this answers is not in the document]
psychocoderHPC: Ok keep it as it is for now. I will have a look into it when I test it with PIConGPU.

bernhardmgruber:
@psychocoderHPC please merge https://github.com/ComputationalRadiationPhysics/mallocMC/pull/179 first! It adds Catch2.

alpaka/include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
bernhardmgruber: I changed this in the alpaka subtree to fix compilation for MSVC and g++. I think it needs to be addressed within alpaka. Here is the issue: https://github.com/alpaka-group/alpaka/issues/1013

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
```diff
@@ -1,59 +0,0 @@
...
-#include <cuda_runtime_api.h>
-#include <string>
-
-namespace mallocMC
-{
-    namespace ReservePoolPolicies
-    {
-        struct CudaSetLimits
-        {
-            static auto setMemPool(size_t memsize) -> void *
```
psychocoderHPC, line 41: I realized that you removed `ReservePoolPolicies` and therefore also the possibility to use a normal malloc on the device. To use malloc on a device you must set configure the heap size.

psychocoderHPC:
I ported the interface change to a prototype branch of PIConGPU https://github.com/psychocoderHPC/picongpu/commit/9edd9e026938f5ddf65a9217df4c2715a3ecf5d3

I will update this prototype again when the `ReservePoolPolicies` are introduced back.

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
bernhardmgruber: It took me a while to understand what you meant, but I think I figured it out: You want to be able to set the creation policy to OldMalloc and the reservePoolPolicy to CudaSetLimits and thus switch the scatter alloc implementation for the default CUDA ::malloc() heap.

I think that should work now, but I am still fighting to add some unit tests for that. But feel free to try it in PIConGPU meanwhile.

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
```diff
@@ -27,26 +27,40 @@
...
+            template<typename AlpakaDev>
+            auto setMemPool(const AlpakaDev & dev, size_t memsize) -> void *
             {
-                cudaDeviceSetLimit(cudaLimitMallocHeapSize, memsize);
+                // FIXME: cannot call cudaDeviceSetLimit() anymore if any kernel has been launched
+                static auto first = cudaDeviceSetLimit(
+                    cudaLimitMallocHeapSize,
+                    memsize); // abusing thread-safe static init for calling
+                              // cudaDeviceSetLimit only once per process
```
bernhardmgruber, line 52: According to CUDA documentation I cannot call this function anymore if a kernel has been run, that used malloc/free. This happens however during the unit tests if you run 2 unit tests, where each one creates an allocator with the `CudaSetLimits` policy.

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
```diff
@@ -27,26 +27,40 @@
...
 
             static void resetMemPool(void * p = nullptr)
             {
-                cudaDeviceSetLimit(cudaLimitMallocHeapSize, 8192U);
+                // see:
+                // https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g05956f16eaa47ef3a4efee84563ccb7d
+                // "Setting cudaLimitMallocHeapSize must not be performed after
+                // launching any kernel that uses the malloc() or free() device
+                // system calls"
+                // cudaDeviceSetLimit(cudaLimitMallocHeapSize, 8192U);
```
bernhardmgruber, line 63: CUDA does not allow us to call this a second time :/ What should we do?

psychocoderHPC: Let us add this behavior to the documentation of the policy `CudaSetLimits` and disable the failing test for this policy. If possible write out a message that the test is disabled for policy XY.

bernhardmgruber: I reverted to the old behavior of the policy and documented the problem. As for the tests, I could not solve them using a `cudaDeviceReset()` alone. It seems I also need to clear the error of the call to `cudaDeviceSetLimit()` in `resetMemPool()`

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
```diff
@@ -47,6 +52,9 @@ namespace mallocMC
             static void resetMemPool(void * p = nullptr)
             {
                 cudaDeviceSetLimit(cudaLimitMallocHeapSize, 8192U);
+                cudaGetLastError(); // cudaDeviceSetLimit() usually fails if any
+                                    // kernel before used ::malloc(), so let's
+                                    // clear the error state
```
bernhardmgruber, line 57: @psychocoderHPC if you are fine with this, and a `cudaDeviceReset()` after each unit test solves all issues.

psychocoderHPC: Yes CUDA device reset is fine but please only in the unit tests so avoid side effects with the user code.

psychocoderHPC:
@bernhardmgruber Could you please integrate the latest alpaka to be able to remove the changes in the alpaka subtree. IMO this is the last part we need to change before we can merge it.

src/include/mallocMC/reservePoolPolicies/CudaSetLimits_impl.hpp
bernhardmgruber: Good. Then we are good already ;)

psychocoderHPC:
big thanks for this PR!!

ax3l:
uff, you just copied in Alpaka under your user name ;-)

nevertheless, thank you for this, great work!

PR #177, opened 2020-05-20
fix used cmake version

psychocoderHPC:
Prefere self installed cmake over the system integrated version.


psychocoderHPC:
this PR is equal to #178. There is a bug in the travis script therefore the CI has not started

PR #178, opened 2020-05-20
fix preferred cmake installation

psychocoderHPC:
Prefere self installed cmake over the system integrated version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
   global:
     - INSTALL_DIR=~/mylibs
     - CXXFLAGS="-Werror"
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
For some reason travis is not reporting back the status and started the CI job. Do not wonder why I have two identical PR's open

psychocoderHPC:
> For some reason travis is not reporting back the status and started the CI job. Do not wonder why I have two identical PR's open

It was a bug in my travis changes

psychocoderHPC:
it should work now: https://travis-ci.org/github/ComputationalRadiationPhysics/mallocMC/builds/689242583#L1136

PR #182, opened 2020-06-26
fix HIP support

psychocoderHPC:
- set warpsize for HIP AMD devices to 64
- fix wrong type for `activemask` - is 64bit if HIP is used
- handle compile issue with HIP-clang

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter_impl.hpp
@@ -619,8 +619,8 @@ namespace mallocMC
                             _page[page].init();
                             // remove chunk information
                             _ptes[page].chunksize = 0;
-#if defined(__CUDA_ARCH__) \
-    || (defined(__HIP_DEVICE_COMPILE__) && defined(__HIP__))
+
+#if(MALLOCMC_DEVICE_COMPILE)
                             __threadfence(); // TODO alpaka
 #else
                             std::atomic_thread_fence(
@@ -782,10 +782,12 @@ namespace mallocMC
 
                 // only one thread per warp can acquire the mutex
                 void * res = 0;
-                const uint32_t mask = activemask();
-                const uint64_t num = popc(mask);
-                const uint64_t lanemask = lanemask_lt();
-                const auto local_id = popc(lanemask & mask);
+                // based on the alpaka backend the lanemask type can be 64bit
+                const auto mask = activemask();
+                const uint32_t num = popc(mask);
+                // based on the alpaka backend the lanemask type can be 64bit
+                const auto lanemask = lanemask_lt();
+                const uint32_t local_id = popc(lanemask & mask);
                 for(unsigned int active = 0; active < num; ++active)
                     if(active == local_id)
                         res = allocPageBasedSingle(acc, bytes);
@@ -807,8 +809,8 @@ namespace mallocMC
             {
                 const uint32 pages = divup(bytes, pagesize);
                 for(uint32 p = page; p < page + pages; ++p) _page[p].init();
-#if defined(__CUDA_ARCH__) \
-    || (defined(__HIP_DEVICE_COMPILE__) && defined(__HIP__))
+
+#if(MALLOCMC_DEVICE_COMPILE)
                 __threadfence(); // TODO alpaka
 #else
                 std::atomic_thread_fence(
@@ -829,7 +831,12 @@ namespace mallocMC
              * @return pointer to the allocated memory
              */
             template<typename AlpakaAcc>
-            ALPAKA_FN_ACC auto create(const AlpakaAcc & acc, uint32 bytes)
+            ALPAKA_FN_ACC
+#if(BOOST_COMP_CLANG && BOOST_LANG_HIP)
+            // The clang HIP compiler is producing runtime errors if this method is inlined.
+            __noinline__
+#endif
+            auto create(const AlpakaAcc & acc, uint32 bytes)
                 -> void *
             {
                 if(bytes == 0)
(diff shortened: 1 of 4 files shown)


alpaka/include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
```diff
@@ -100,7 +100,7 @@ namespace alpaka
                     std::int64_t value)
                 -> std::int32_t
                 {
-                    return __ffsll(value);
+                    return __ffsll(static_cast<unsigned long long>(value));
```
bernhardmgruber, line 103: This might work for HIP, but `__ffsll` in CUDA takes a `long long int` as argument, so now there is a narrowing conversion in CUDA.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -782,10 +785,12 @@ namespace mallocMC
...
                 // only one thread per warp can acquire the mutex
                 void * res = 0;
-                const uint32_t mask = activemask();
-                const uint64_t num = popc(mask);
-                const uint64_t lanemask = lanemask_lt();
-                const auto local_id = popc(lanemask & mask);
+                // based on the alpaka backend the mask type can be 64bit
+                const auto mask = activemask();
+                const uint32_t num = popc(mask);
+                // based on the alpaka backend the mask type can be 64bit
```
bernhardmgruber, line 791: the *lanemask* type

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -1019,10 +1024,15 @@ namespace mallocMC
...
+                auto threadsPerBlock = VecType::ones();
+                threadsPerBlock[Dim::value - 1] = 256u;
+
                 const auto workDiv = alpaka::workdiv::WorkDivMembers<Dim, Idx>{
-                    Idx{1},
-                    Idx{256},
-                    Idx{1}}; // Dim may be any dimension, but workDiv is 1D
+                    VecType::ones(),
+                    threadsPerBlock,
+                    VecType::ones()}; // Dim may be any dimension, but workDiv is 1D
```
bernhardmgruber, line 1035: What is the purpose of this change? The original version should have the same behavior and is shorter.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -1218,10 +1228,18 @@ namespace mallocMC
...
+                auto threadsPerBlock = VecType::ones();
+                threadsPerBlock[Dim::value - 1] = 256u;
+
                 const auto workDiv = alpaka::workdiv::WorkDivMembers<Dim, Idx>{
-                    Idx{64},
-                    Idx{256},
-                    Idx{1}}; // Dim may be any dimension, but workDiv is 1D
+                    numBlocks,
+                    threadsPerBlock,
+                    VecType::ones()}; // Dim may be any dimension, but workDiv is 1D
```
bernhardmgruber, line 1242: Same here (see above).

alpaka/include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
psychocoderHPC: HIP also provide a `long long` version I will fix it more clean in https://github.com/alpaka-group/alpaka/pull/1049/files

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
psychocoderHPC: The original version produces compile time errors because of https://github.com/alpaka-group/alpaka/issues/1035
The changes setting always the x component (highest) index depending of the user provided accelerator.

psychocoderHPC:
@bernhardmgruber could you please review this PR again. I updated the PR and removed changes in the alpaka subtree.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
bernhardmgruber: Hmm. But that did not happen with the alpaka subtree? Then we should probably also update the alpaka subtree. Did we not release an alpaka version recently? We could update the subtree to this version.

As for the indexing, that once worked independently on the dimension of the accelerator. But maybe this was more of a bug than a feature. Anyway, if it works now as well, I am fine with it!

src/include/mallocMC/mallocMC_utils.hpp
```diff
@@ -123,18 +132,33 @@ namespace mallocMC
...
 
+    ALPAKA_FN_ACC inline auto ballot(int pred)
+    {
+#if defined(__CUDA_ARCH__)
+        return __ballot_sync(__activemask(), pred);
+#elif(MALLOCMC_DEVICE_COMPILE_PATH && BOOST_COMP_HIP)
+        // return value is 64bit for HIP-clang
+        return __ballot(pred);
+#else
+        return 1u;
```
bernhardmgruber, line 150: I think we should `return pred;` here.
On CPU implementations, where there is only 1 thread voting, `ballot()` should return what the 1 thread put in as predicate.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
```diff
@@ -619,8 +619,8 @@ namespace mallocMC
                             _page[page].init();
                             // remove chunk information
                             _ptes[page].chunksize = 0;
-#if defined(__CUDA_ARCH__) \
-    || (defined(__HIP_DEVICE_COMPILE__) && defined(__HIP__))
+
+#if(MALLOCMC_DEVICE_COMPILE_PATH)
```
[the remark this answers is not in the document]
psychocoderHPC: No not required in [cupla](https://github.com/alpaka-group/cupla/blob/f2cd8e551ebbe10c1a2c877ed8f07ee6c0e84288/include/cupla/defines.hpp#L124) I will change it.

src/include/mallocMC/mallocMC_utils.hpp
psychocoderHPC: ballot is creating a bitmask where the bit of the thread is set where `pred` is non zero. `pred` can be any integer value.

From the [cuda documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-vote-functions)
```
Evaluate predicate for all non-exited threads in mask and return an integer whose Nth bit is set if and only if predicate evaluates to non-zero for the Nth thread
```

psychocoderHPC: As soon as alpaka is providing the intrinsic we will remove this code anyway.

src/include/mallocMC/creationPolicies/Scatter_impl.hpp
psychocoderHPC: It was with the alpaka subtree and if I remember correctly with HIP-clang (clang 10).
I will keep it for now as it is, because it is save and open a issue that we invest later time into it.

PR #181, opened 2020-06-23
Refactoring

bernhardmgruber:
* made struct with a constant just a constant
* refactored iostream usage
* merged all *_impl headers into their declaring headers
* including mallocMC.hpp instead of individual headers in tests and examples

PR #196, opened 2020-11-09
remove workround for HIP

psychocoderHPC:
Workaround was introduces with #181 and is not valid anymore.
HIP3.8+ requires that this line is removed.


bernhardmgruber:
Thank you for revisiting such old workarounds and removing them, if they are no longer required!

bernhardmgruber (approved):
I cannot verify your changes, but I approve removing workarounds if they are no longer necessary.

sbastrakov (approved):

PR #197, opened 2020-11-09
Update .clang-format and apply clang-format

bernhardmgruber:
.clang-format file taken from here: https://github.com/ComputationalRadiationPhysics/contributing

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,64 +1,79 @@
 ---
+# General options
+Language: Cpp
+Standard: c++17
+DisableFormat: false
+
 AccessModifierOffset: -4
 AlignAfterOpenBracket: AlwaysBreak
 AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
 AlignConsecutiveDeclarations: false
-AlignEscapedNewlines: DontAlign
-AlignOperands: false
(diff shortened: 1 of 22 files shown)


sbastrakov (commented):
Maybe it's better if this PR is split into two commits: one updates only the `.clang-format` file and is done normally. While the second adds the result of its application and is "authored" by some artificial account. I believe this is what @psychocoderHPC meant on the last alpaka meeting. For an example, see [this one](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3244#issue-419976879), also I think the command line in PR description is helpful to someone else re-doing this step later.

bernhardmgruber:
You are absolutely right! I forgot about that! Thanks!

psychocoderHPC:
As @sbastrakov wrote please add a separate commit for the clang file that we have a diff to the old and add the commands used to the PR description. No need to have the commands in the `commit descroption`.

bernhardmgruber:
Here is a shell command I put in CONTRIBUTION.md half a year ago: https://github.com/alpaka-group/mallocMC/blob/dev/CONTRIBUTING.md

psychocoderHPC (approved):

PR #204, opened 2021-01-21
HIP: get warpsize from define

psychocoderHPC:
For HIP 4.0+ use `__AMDGCN_WAVEFRONT_SIZE` to get the warp size.


bernhardmgruber:
I have not tested it, but the change seems legit.

What is more surprising is that querying the warpsize is not inside alpaka. I feel this functionality should be there.

bernhardmgruber:
After a bit of googling and browsing through the documentation, I think we need the extra constant because the CUDA and HIP `warpSize` is a variable and not a constant.

bernhardmgruber (approved):

PR #206, opened 2021-01-21
switch from rodare to zenodo

psychocoderHPC:
Old releases were already published on zenodo so we keep zenodo instead of using HZDR rodare.
To keep  zenodo was requested by @ax3l [here](https://github.com/alpaka-group/mallocMC/pull/200#issuecomment-760559982).

This PR is against the release branch 2.5.0 and will be backported after we released.


.zenodo.json
```diff
@@ -0,0 +1,50 @@
+{
+  "title": "mallocMC - Memory Allocator for Many Core Architectures",
+  "description": "This project provides a framework for fast memory managers on many core accelerators. It is based on alpaka to run on many different accelerators and implements the ScatterAlloc algorithm.",
+  "keywords": [
+    "mallocMC",
+    "CUDA",
+    "manycore",
+    "GPU"
```
bernhardmgruber, line 8: I would add `allocator` her as well.

bernhardmgruber (approved):
I tried to find a schema for these zenodo files, but I could not find one. The file is at least valid JSON. I trust you that it works. Otherwise just send another PR.

psychocoderHPC:
> I tried to find a schema for these zenodo files, but I could not find one. The file is at least valid JSON. I trust you that it works. Otherwise just send another PR.

I also past it always into a validator. A rodare validator would be nice.

ax3l:
@bernhardmgruber the json schema is the same _invenio_ schema

bernhardmgruber:
Thank you @ax3l! It seems you can get the JSON schema there somewhere, but it seems there is no userfriendly online validator available. Anyway, thx!

ax3l:
Yep, would like that too.
Maybe one can write a gh action that uses the invenio schema python package to validate it on changes or so.

bussmann:
Just my 2 cents: Helmholtz is working on key performance indices for Software. Then your rodare software can be easily added into the reporting (which is A LOT of work). Integration Zenodo: Not so much. Btw: I think KPIs for Software are stupid!

psychocoderHPC:
@bernhardmgruber ready to merge

.zenodo.json
```diff
@@ -0,0 +1,51 @@
...
+    {
+      "name":"Worpitz, Benjamin"
+    },
+    {
+      "name":"Grund, Alexander",
+      "affiliation":"Helmholtz-Zentrum Dresden-Rossendorf",
+      "orcid":"0000-0002-7196-8452"
+    }
+  ]
+}
```
sbastrakov, line 51: @psychocoderHPC is this missing new line important?

psychocoderHPC: I will add it to be sure that it is not making trouble

sbastrakov (approved):

Issue #212, opened 2021-02-18
XMallocSIMD broken for CPU


psychocoderHPC:
The implementation of the distribution policy `XMallocSIMD` is broken for CPU accelerators.

The reason is that we use `alpaka::declareSharedVar` within a function that can be called multiple times by the users.

see: https://github.com/alpaka-group/alpaka/issues/1257

psychocoderHPC:
This is already fixed in alpaka.

PR #213, opened 2021-02-18
add full support for CPU accelerators

psychocoderHPC:
Until now we supported only alpaka accelerators which can be used with
multiple threads per block.

- add support for more accelerators in the example (except TBB and BoostFiber)
- remove usage of the distribution policy `XMallocSIMD` because of #212 (this will be reverted after the issue if fixed)

### DIFF
--- a/examples/mallocMC_example01.cpp
+++ b/examples/mallocMC_example01.cpp
@@ -26,16 +26,19 @@
   THE SOFTWARE.
 */
 
+#include <algorithm>
 #include <alpaka/alpaka.hpp>
+#include <alpaka/example/ExampleDefaultAcc.hpp>
+#include <cassert>
 #include <iostream>
 #include <mallocMC/mallocMC.hpp>
 #include <numeric>
 
 using Dim = alpaka::DimInt<1>;
 using Idx = std::size_t;
-// using Acc = alpaka::AccCpuThreads<Dim, Idx>;
-// using Acc = alpaka::AccCpuOmp2Threads<Dim, Idx>;
-using Acc = alpaka::AccGpuCudaRt<Dim, Idx>;
+
+// Define the device accelerator
+using Acc = alpaka::ExampleDefaultAcc<Dim, Idx>;
 
 struct ScatterHeapConfig
 {
@@ -67,7 +70,7 @@ struct ShrinkConfig
 using ScatterAllocator = mallocMC::Allocator<
     Acc,
     mallocMC::CreationPolicies::Scatter<ScatterHeapConfig, ScatterHashConfig>,
-    mallocMC::DistributionPolicies::XMallocSIMD<XMallocConfig>,
+    mallocMC::DistributionPolicies::Noop,
     mallocMC::OOMPolicies::ReturnNull,
     mallocMC::ReservePoolPolicies::AlpakaBuf<Acc>,
     mallocMC::AlignmentPolicies::Shrink<ShrinkConfig>>;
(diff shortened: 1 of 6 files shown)


examples/mallocMC_example01.cpp
```diff
@@ -78,14 +93,20 @@ ALPAKA_STATIC_ACC_MEM_GLOBAL int** arC;
...
     constexpr auto length = 100;
-    static_assert(length <= block * grid, ""); // necessary for used algorithm
 
     const auto dev = alpaka::getDevByIdx<Acc>(0);
     auto queue = alpaka::Queue<Acc, alpaka::Blocking>{dev};
 
+    auto const devProps = alpaka::getAccDevProps<Acc>(dev);
+    unsigned block = std::min(static_cast<size_t>(32u), static_cast<size_t>(devProps.m_blockThreadCountMax));
+
+    std::cout<< block<<std::endl;
```
bernhardmgruber: Is this correctly formatted?

examples/mallocMC_example01.cpp
```diff
@@ -30,12 +30,27 @@
...
+using Acc = alpaka::AccCpuThreads<Dim, Idx>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED)
+using Acc = alpaka::AccCpuOmp2Threads<Dim, Idx>;
+#elif defined(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED)
+using Acc = alpaka::AccCpuOmp2Blocks<Dim, Idx>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+using Acc = alpaka::AccCpuSerial<Dim, Idx>;
+#else
+#error "Selected alpaka accelerator is not supported!"
+#endif
```
bernhardmgruber: This looks similar to the alpaka example accelerator. Could we use this one instead?

examples/mallocMC_example01.cpp
```diff
@@ -67,7 +82,7 @@ struct ShrinkConfig
 using ScatterAllocator = mallocMC::Allocator<
     Acc,
     mallocMC::CreationPolicies::Scatter<ScatterHeapConfig, ScatterHashConfig>,
-    mallocMC::DistributionPolicies::XMallocSIMD<XMallocConfig>,
+    mallocMC::DistributionPolicies::Noop,
```
bernhardmgruber, line 73: Is this PR urgent? Because I would prefer fixing `declareSharedVar` in alpaka and not disable it temporarily.

examples/mallocMC_example03.cpp
```diff
@@ -100,10 +114,12 @@ auto main() -> int
 {
     const auto dev = alpaka::getDevByIdx<Acc>(0);
     auto queue = alpaka::Queue<Acc, alpaka::Blocking>{dev};
+    auto const devProps = alpaka::getAccDevProps<Acc>(dev);
+    unsigned block = std::min(static_cast<size_t>(32u), static_cast<size_t>(devProps.m_blockThreadCountMax));
```
bernhardmgruber: I think `block` can be `const`.

tests/verify_heap.cpp
```diff
@@ -614,12 +614,17 @@ void print_machine_readable(
...
+    unsigned threads,
     const bool machine_readable) -> bool
 {
     const auto dev = alpaka::getDevByIdx<Acc>(0);
     auto queue = Queue{dev};
 
+    auto const devProps = alpaka::getAccDevProps<Acc>(dev);
+    unsigned maxThreads = static_cast<size_t>(devProps.m_blockThreadCountMax);
+    if(maxThreads < threads)
+        threads = maxThreads;
```
bernhardmgruber: I would prefer: `threads = std::min(threads, static_cast<size_t>(devProps.m_blockThreadCountMax));`

examples/mallocMC_example01.cpp
psychocoderHPC: well spotted, it is a leftover from my tests

examples/mallocMC_example01.cpp
psychocoderHPC: I will enable the usage as soon as it is fixed in alpaka, currently, mallocMC would crash with CPU accelerators.

bernhardmgruber: You mean reenable in a separate PR?

psychocoderHPC: Yes I will re-enable it in a separate PR after it is fixed in alpaka.

bernhardmgruber: Go ahead then.

examples/mallocMC_example01.cpp
bernhardmgruber: @psychocoderHPC: Can we use `alpaka::ExampleDefaultAcc<Dim, Idx>` here?

bernhardmgruber (approved):

examples/mallocMC_example01.cpp
bernhardmgruber: I think I would have prefered a TODO comment here, but I trust that you will change it back once the issue with `declareSharedVar` is fixed in alpaka.

bernhardmgruber (approved):

PR #216, opened 2021-02-22
fix clang format test

psychocoderHPC:
exclude alpaka


sbastrakov (approved):

bernhardmgruber:
Is there a broader rational for this change? While it is fine that we only clang-format the mallocMC sources, I wonder if you encountered any other issues? Alpaka should be formated well as well, so there should never be an issue without your PR.

PR #218, opened 2021-10-05
fix scatter allocation

psychocoderHPC:
In case pagesize bytes will be allocated and within `allocChunked` an
allocation is increased based on the waste factor to fit into a chunk
with the size of a page the locking is not working because `allocPageBased`
locking is not compatible to the locking mechanism used in
`allocChunked`.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -767,10 +767,13 @@ namespace mallocMC
             {
                 if(bytes == 0)
                     return 0;
-                // take care of padding
-                // bytes = (bytes + dataAlignment - 1) & ~(dataAlignment-1); //
-                // in alignment-policy
-                if(bytes < pagesize)
+                /* Take care of padding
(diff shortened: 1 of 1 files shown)


src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -767,10 +767,13 @@ namespace mallocMC
             {
                 if(bytes == 0)
                     return 0;
-                // take care of padding
-                // bytes = (bytes + dataAlignment - 1) & ~(dataAlignment-1); //
-                // in alignment-policy
-                if(bytes < pagesize)
+                /* Take care of padding
+                 * bytes = (bytes + dataAlignment - 1) & ~(dataAlignment-1);
+                 * in alignment-policy.
```
sbastrakov, line 772: Are these 3 lines relevant to this part? I know this PR doesn't change it, but probably they belong somewhere else?

psychocoderHPC: I kept the equation separated from the normal text that it is better readable.

sbastrakov (approved):

PR #221, opened 2021-10-05
fix `calcAdditionalChunks`

psychocoderHPC:
Additional chunks can not be more than 31. In case the page is large and the allocation small the current implementation could return wrong results larger 31.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -330,11 +330,14 @@ namespace mallocMC
                 uint32 chunksize) -> uint32
             {
                 if(fullsegments != 32)
-                    return alpaka::math::max(
-                               acc,
-                               0u,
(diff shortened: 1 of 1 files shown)


src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -330,11 +330,14 @@ namespace mallocMC
...
             {
                 if(fullsegments != 32)
-                    return alpaka::math::max(
-                               acc,
-                               0u,
-                               (int) pagesize - (int) fullsegments * segmentsize - (int) sizeof(uint32))
-                        / chunksize;
+                    return alpaka::math::min(
+                        acc,
+                        31,
```
sbastrakov, line 335: Could this `31` (and/or the `32` above) become a constexpr variable with some telling name?

psychocoderHPC: I will do it in separate PR.

sbastrakov (approved):

PR #224, opened 2021-10-07
fix: Scatter is not checking all regions

psychocoderHPC:
bug introduced with #222

The bug is fixed by a rewrite of the walk-over pages in `allocChunked()`.
The new method to walk over pages additionally solved a bug that is already in the original scatter implementation.
`accessblocks` is only a virtual concept and the old method to move over pages do not take into account that the number of pages is not always a multiple of regions per access block.
The rewrite simplified the code and reduced the register footprint by 1-2 registers.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -487,10 +487,11 @@ namespace mallocMC
             {
                 // use the minimal allocation size to increase the hit rate for small allocations.
                 const uint32 minAllocation = alpaka::math::max(acc, bytes, +minChunkSize);
-                const uint32 pagesperblock = _numpages / accessblocks;
+                const uint32 numpages = _numpages;
+                const uint32 pagesperblock = numpages / accessblocks;
                 const uint32 reloff = warpSize * minAllocation / pagesize;
-                const uint32 startpage = (minAllocation * hashingK + hashingDistMP * smid()
-                                          + (hashingDistWP + hashingDistWPRel * reloff) * warpid())
+                const uint32 start_page_in_block = (minAllocation * hashingK + hashingDistMP * smid()
+                                                    + (hashingDistWP + hashingDistWPRel * reloff) * warpid())
                     % pagesperblock;
                 const uint32 maxchunksize = alpaka::math::min(
                     acc,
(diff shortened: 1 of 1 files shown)


src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -500,8 +501,8 @@ namespace mallocMC
                      */
                     alpaka::math::max(acc, wastefactor * bytes, +minChunkSize));
 
-                const uint32 startblock = _firstfreeblock;
-                const uint32 random_start_page = startpage + startblock * pagesperblock;
+                // global page index
+                const uint32 random_start_page = start_page_in_block + _firstfreeblock * pagesperblock;
```
sbastrakov:
```suggestion
                const uint32 global_start_page = start_page_in_block + _firstfreeblock * pagesperblock;
```

I think global is better fitting in the name, since this function already use this kind of naming for some other variables. And the fact it is randomly choosen doesn't mattter for the rest of the code, it is only important that it is the fixed start page

psychocoderHPC: changed

src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -515,69 +519,73 @@ namespace mallocMC
                  */
                 for(uint32 finder = 0; finder < 2; ++finder)
                 {
-                    uint32 start_page = random_start_page;
-                    uint32 b = startblock;
+                    uint32 global_page = global_start_page;
+                    // loop over pages until we arrive global_start_page again
                     do
```
sbastrakov:
```suggestion
                    /* Loop over all pages until we found a free one or arrived to global_start_page again
                     * This and the following loop are done as do-while to potentially save registers by avoiding an extra loop counter variable
                     */
                    do
```

src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -515,69 +519,76 @@ namespace mallocMC
...
+                            // loop over pages within a region
+                            do
                             {
-                                uint32 selected_page = start_page;
-                                do
+                                const uint32 chunksize = _ptes[page_in_region].chunksize;
+                                if(chunksize >= bytes && chunksize <= maxchunksize)
+                                {
+                                    void* res = tryUsePage(acc, page_in_region, chunksize);
+                                    if(res != 0)
```
sbastrakov:
```suggestion
                                    if(res != nullptr)
```

sbastrakov: Also in same conditions below

src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -515,69 +519,76 @@ namespace mallocMC
...
-                                    alpaka::atomicOp<alpaka::AtomicCas>(
-                                        acc,
-                                        (uint32*) (_regions + region),
-                                        regionfilllevel,
-                                        regionfilllevel + 1);
-                            }
-                            start_page += regionsize;
+                                }
+                                page_in_region = region_offset + ((page_in_region + 1) % regionsize);
+                            } while(global_page != page_in_region);
```
sbastrakov:
```suggestion
                            } while(page_in_region != global_page);
```
I think this is more natural as `page_in_region` is the loop counter here

sbastrakov (approved):

Issue #227, opened 2021-10-27
Some recent change broke alignment


sbastrakov:
PIConGPU recently updated to `dev` branch of mallocMC and started reliably crashing in 2d simulations: https://github.com/ComputationalRadiationPhysics/picongpu/issues/3891 . I've narrowed down the crash to this update of mallocMC inside PIConGPU. So probably one of the recent PRs #217 - #226 introduced some bug that's causing it. The fact it's 2d in PIConGPU may be a conincidence or just one situation it appears.

We need to figure out and solve it quickly, as current PIConGPU `dev` and the release candidate branch are now impossible (or unreliable) to use for 2d because of this bug.

sbastrakov:
Investigating that PIConGPU crash makes #220 suspect. I manually changed mallocMC inside PIConGPU to after all PRs in the suspect range, that case was working before #220 was merged, but not after.

cc @bernhardmgruber , beware this mallocMC version.

PR #220, opened 2021-10-05
Scatter: optimize hash and allocation size

psychocoderHPC:
- calculate `minChunkSize` based on the limitations of scatter alloc
- allocations smaller than `minChunkSize` will always allocate `minChunkSize` bytes
- use optimized allocation size for the hash function to increase the possibility to find a usable slot

PR #228, opened 2021-10-28
Add alignment to minChunkSize in ScatterAlloc

sbastrakov:
After the changes of #220 the value of `minChunkSize` is generally not a power of 2. E.g. for 2 MB pages it was equal to 2047. For all allocations smaller than that, `ScatterAlloc` will allocate `minChunkSize` bytes instead in its `allocChunked()`.

I believe something is wrong there, and it causes #227. With the fix in this PR my previously failing PIConGPU setup now works. However, I am not sure what should the "alignment" be. Chose 16 as that was the value of `minChunkSize` before #220. @psychocoderHPC you probably know better what to use here.

We need to have this fix urgently, as it has to be ported to PIConGPU `dev` and release candidate branches.

Solves #227 .

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -175,7 +175,11 @@ namespace mallocMC
              * existing chunk.
              * Each page can have 32x32 chunks. To maintain 32 chunks we need 32 bitmask on the page (each 32bit)
              */
-            static constexpr uint32 minChunkSize = (pagesize - 32u * sizeof(uint32)) / (32u * 32u);
+            static constexpr uint32 minChunkSizeUnaligned = (pagesize - 32u * sizeof(uint32)) / (32u * 32u);
+            // Set minChunkSize to largest multiple of minChunkSizeAlignment that is <= minChunkSizeUnaligned
+            static constexpr uint32 minChunkSizeAlignment = 16u;
+            static constexpr uint32 minChunkSize
+                = (minChunkSizeUnaligned / minChunkSizeAlignment) * minChunkSizeAlignment;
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
`minChunksize` is not required to be a power of two because `chunksize != alignment`. mallocMC has a policy for alignment. Chunk size is used to group allocations only.

src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -175,7 +175,11 @@ namespace mallocMC
              * existing chunk.
              * Each page can have 32x32 chunks. To maintain 32 chunks we need 32 bitmask on the page (each 32bit)
              */
-            static constexpr uint32 minChunkSize = (pagesize - 32u * sizeof(uint32)) / (32u * 32u);
+            static constexpr uint32 minChunkSizeUnaligned = (pagesize - 32u * sizeof(uint32)) / (32u * 32u);
+            // Set minChunkSize to largest multiple of minChunkSizeAlignment that is <= minChunkSizeUnaligned
+            static constexpr uint32 minChunkSizeAlignment = 16u;
+            static constexpr uint32 minChunkSize
```
psychocoderHPC, line 181: if you like to align you need to round up.

```
static constexpr uint32 minChunkSize
   = ((minChunkSizeUnaligned + minChunkSizeAlignment -1u)/ minChunkSizeAlignment) * minChunkSizeAlignment;
```

sbastrakov: I think that would be wrong as then the data might not fit the page. Probably easier to discuss on a VC

sbastrakov:
This PR tried the fix the correctly identified problem, but the fix is generally wrong. #229 provides an actual fix.

PR #229, opened 2021-11-03
fix wrong alignment

### LINKED ISSUE #227: Some recent change broke alignment

psychocoderHPC:
fix #227 and is a replacement for #228

In cases where the data allocation size is smaller than the minimum chunk size, the allocation size will be set to minimal chunk size.
In such cases, the alignment of the data can be wrong.

mallocMC inerface changes (not visible for the user):
- CreationPolicy::create() get the alignment policy as first template parameter
- AlignmentPolicy::applyPadding() requires 32bit input sizes and returns the size as 32bit value. (avoid cast operations)

Note: The changes will not increase the register-footprint (tested with PIConGPU)

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -174,6 +174,8 @@ namespace mallocMC
              * fragment the heap and increases the possibility that a small allocation can reuse an
              * existing chunk.
              * Each page can have 32x32 chunks. To maintain 32 chunks we need 32 bitmask on the page (each 32bit)
+             *
+             * @note: There is no requirement that minChunksSize is a power of two.
              */
             static constexpr uint32 minChunkSize = (pagesize - 32u * sizeof(uint32)) / (32u * 32u);
             static constexpr uint32 minSegmentSize = 32u * minChunkSize + sizeof(uint32);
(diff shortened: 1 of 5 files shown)


src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -174,6 +174,8 @@ namespace mallocMC
              * fragment the heap and increases the possibility that a small allocation can reuse an
              * existing chunk.
              * Each page can have 32x32 chunks. To maintain 32 chunks we need 32 bitmask on the page (each 32bit)
+             *
+             * Attention: There is no requirement that minChunksSize is a power of two.
```
sbastrakov: It's more of a `@note` than attention, but no big deal

sbastrakov (approved):

sbastrakov:
I did testing with PIConGPU, with this PR the issue https://github.com/ComputationalRadiationPhysics/picongpu/issues/3891 disappeared.

PR #241, opened 2023-07-12
fix creation policy Scatter

psychocoderHPC:
It is possible that mallocMC is giving the same memory slot to different concurrent allocation calls.
The error shows up in very seldom cases in PIConGPU where many threads on CUDA Ampere GPUs allocated memory of different sizes at the same time.

https://github.com/ComputationalRadiationPhysics/picongpu/issues/4576#issuecomment-1632575759

The data race happens because the last thread is emptying the PTE with `_page[page].init();`. Most likely the data race happened because we alter the bitmasks with atomics but reset the bitmask in init without. A second potential reason could be that we read the chunk size in one place without using atomic which is resulting into data races too.

This PR
- removes reading chunk size without atomic
- is changing the way how a PTE is freed
- reduce branch divergence by refactoring the calls to `tryUsePage()`

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -206,13 +206,18 @@ namespace mallocMC
 #endif
             static constexpr uint32 hashingDistWPRel = MALLOCMC_CP_SCATTER_HASHINGDISTWPREL;
 
-            /**
-             * Page Table Entry struct
+            /** Page Table Entry struct
+             *
              * The PTE holds basic information about each page
              */
             struct PTE
             {
                 uint32 chunksize;
+                /** Counter for how many page table entries are used.
+                 *
+                 * This counter is used internally as lock, to guard a full PTE the value must be set to pagesize via
+                 * atomic CAS.
+                 */
                 uint32 count;
                 uint32 bitmask;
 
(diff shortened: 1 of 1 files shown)


pordyna:
I ran the previously crushing setup on `perlmutter` with this fix in a loop, for approximately 4h, and didn't crash.

src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -652,30 +670,56 @@ namespace mallocMC
...
-                            // unlock it
-                            alpaka::atomicOp<alpaka::AtomicSub>(acc, (uint32*) &_ptes[page].count, +pagesize);
-                        }
+                        // clean meta data bits on the PTE
+                        _page[page].init();
+
+                        // remove chunk information
+                        alpaka::atomicOp<alpaka::AtomicCas>(acc, (uint32*) &_ptes[page].chunksize, chunksize, 0u);
+
+                        /** Take care that the meta data changes where we do not used atomic are propagated to all
```
pordyna:
```suggestion
                        /** Take care that the meta data changes where we did not use atomics are propagated to all
```

pordyna (approved):
It looks good to me, as far as I can understand what is happening here (and I don't understand much^^).

But, just for getting some idea about what happens ;) , there were 3 changes? 
* Disabling caching with volatile when writing zeros
*  this refactoring in `tryUsePage`
* and we are now locking first and reducing fill level afterwards so that it should be always the same thread that locks/clears meta and reduces the fill counter? And we were doing it the other way around before, and this is somehow better.

Did I miss sth?

psychocoderHPC:
> * and we are now locking first and reducing fill level afterwards so that it should be always the same thread that locks/clears meta and reduces the fill counter? And we were doing it the other way around before, and this is somehow better.

During freeing memory we try to lock the page but this only succeeds in case we hold the last allocation on the page. 
In any case we reduce afterward the fill level, by one in case locking was not possible and by `pagesize` in case locking before works.
Before this PR we reduced the fill level first and afterward we tried to lock the page in case we was the last allocation on the page. In general, this should work too but was not :-(

PR #247, opened 2024-03-26
Fit `accessblocksize` into heap for examples and tests

chillenzer:
As realised by @psychocoderHPC, `accessblocksize` didn't fit into the total allocation size such that no pages were actually created. Now, examples are runnable again.


chillenzer:
Added an identical fix for `tests/`. This does not (or only partially) solve #249.

chillenzer:
@psychocoderHPC should we include this into the upcoming release or close?

psychocoderHPC (approved):

PR #253, opened 2024-06-17
fix few bugs

psychocoderHPC:
- fix double page clean-up()` was not taken into account
- fix that the page lock during `tryUsePage

@chillenzer could you please have a look if you observe any logical issues.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -463,12 +463,12 @@ namespace mallocMC
                 // increse the fill level
                 const uint32 filllevel = alpaka::atomicOp<alpaka::AtomicAdd>(acc, (uint32*) &(_ptes[page].count), 1u);
 
-                // if resetfreedpages == false we do not need to re-check filllevel or chunksize
+                // if resetfreedpages == false we do not need to re-check chunksize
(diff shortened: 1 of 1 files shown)


chillenzer (approved):
Looks okay to me. In contrast to my version, you've only got 2 calls to `tryCleanPage` when it's 4 in my case. One of them is when trying out pages. Because you never "enter" a page (increasing filling level) while searching around, you don't take responsibility for cleaning up. The other one is the multi-page mode which does this by hand.

Apart form that, the structure seems to be very similar.

psychocoderHPC:
> Looks okay to me. In contrast to my version, you've only got 2 calls to `tryCleanPage` when it's 4 in my case. One of them is when trying out pages. Because you never "enter" a page (increasing filling level) while searching around, you don't take responsibility for cleaning up. The other one is the multi-page mode which does this by hand.
> 
> Apart form that, the structure seems to be very similar.

I will check tomorrow for the cases where the full page is used. Since in this implementation chunksize is checked before increasing the fill level the number of calls to try clean can differ.

psychocoderHPC:
I checked if `tryCleanPage` is called where it should The multi page allocation for large chunks is not using the fill level. It uses the chunksize only therefore `tryCleanPage` is not required.

PR #254, opened 2024-06-25
Add FlatterScatter creation policy

chillenzer:
This is WIP and the original description here is horribly outdated. I'll keep this in for easy access to the diff but I'll update once it's relevant.
Hi @psychocoderHPC,

this version has a compile-time defined heap. Should theoretically work as almost drop-in replacement for the old version with the following changes:

- `wastefactor == 1` and `resetfreedpages == true` are enforced at compile-time.
- The `ScatterConfig` struct used to configure the heap needs one more compile-time value `heapsize`.
- There is a runtime value for `heapsize` given to the constructor of the `Allocator`. The two are enforced to be identical.

Feel free to play around with it!


src/include/mallocMC/creationPolicies/Scatter.hpp
```diff
@@ -33,1342 +27,465 @@
...
-                 * on the waste factor is colliding with the allocation schema in allocPageBased().
-                 */
-                if(bytes <= pagesize)
-                    // chunck based
-                    return allocChunked<AlignmentPolicy>(acc, bytes);
-                else
-                    // allocate a range of pages
-                    return allocPageBased(acc, bytes);
+                leavePage(acc, index);
+                ++index;
```
psychocoderHPC: BUG: must be `index = (index + 1u) % numPages();` else index can become to large

chillenzer: Nope, that's not a problem:
1. `index` can get out of range (slightly, i.e., by 1 if `index` was pointing to the last page before) but that's not a problem because it is only used in a `wrappingLoop`. Any `startIndex` out of range will simply result in effectively starting from 0.
2. `index` is set on the following line to the new hoped-to-be-free page which is again a valid (potentially wrapped around) index in range, so `index` cannot become arbitrarily large.

chillenzer:
Superseeded by #271.

PR #255, opened 2024-08-06
Fix memory order for c++20 compatibility

ikbuibui:


psychocoderHPC:
IMO this PR is breaking C++ < 20 because before 20 it is a enum. 

https://en.cppreference.com/w/cpp/atomic/memory_order

We need to handle c++17 and C++20 and newer via a macro. @ikbuibui could you check if we can use alpaka threadfence methods instead.
I assume this code is pr alpaka thread fence but now we can use alpaka natives.

ikbuibui:
In my tests it works with C++17 and alpaka thread fence uses the same

PR #256, opened 2024-08-21
Fix non chunk size resetting

chillenzer:
Fixes a deadlock popping up applying the test suite for `FlatterScatter` to `Scatter`.

### DIFF
--- a/src/include/mallocMC/creationPolicies/Scatter.hpp
+++ b/src/include/mallocMC/creationPolicies/Scatter.hpp
@@ -978,7 +978,7 @@ namespace mallocMC
                     mem = (void*) counter;
                 }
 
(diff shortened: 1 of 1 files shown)


chillenzer:
Offline discussion: This is not so much a bug as a misimplementation of the `resetfreedpages` feature because it simply means that pages are potentially left with a chunk size while no thread holds memory there. While it can have unfortunate implications concerning memory availability (which led to deadlocking in the stress test code, i.e. user code), it is not an urgent bug in the strict sense or a deadlock in the library code. Setting this to draft as a reminder that we might want to do a major refactoring of the algorithm in order to establish an invariant that can be used to prove correctness then.

PR #271, opened 2024-11-12
FlatterScatter

### LINKED ISSUE #249: Weird test failures
On the system at hand (HAL), the tests are all passing individually but fail if and only if "2D AccGpuCudaRt" and "3D AccGpuCudaRt" are both run in the same test run, so the following works
```bash
./tests <any test name>
./tests -f list.txt
```
whenever `list.txt` doesn't contain both of the above while the latter fails with any other `list.txt`, in particular
```
2D AccGpuCudaRt
3D AccGpuCudaRt
```
as a MWE. From what I can tell, neither the ordering of the two nor other tests in between change this observation. The error message from the `./tests` is always
```
FAILED:
due to unexpected exception with message:
  /home/lenz93/workspace/src/mallocMC-makeReservationPolicyOwning/alpaka/
  include/alpaka/platform/PlatformUniformCudaHipRt.hpp(125) 'rc' A previous API
  call (not this one) set the error  : 'cudaErrorMisalignedAddress':
  'misaligned address'!
```
while
```bash
$ compute-sanitizer --show-backtrace=yes  --tool memcheck ./tests
========= COMPUTE-SANITIZER
========= Invalid __global__ atomic of size 4 bytes
=========     at atomicSub(unsigned int *, unsigned int)+0x4c0 in /opt/spack-modules/opt/spack/linux-ubuntu22.04-zen2/gcc-12.2.0/cuda-12.3.0-vhkdpnlaplsol7vzlt4wj55im3uemjkq/include/device_atomic_functions.hpp:122
=========     by thread (0,1,0) in block (0,3,1)
=========     Address 0x7ff15fe22051 is misaligned
=========     and is inside the nearest allocation at 0x7ff15fe00000 of size 1,048,576 bytes
=========     Device Frame:alpakaGlobal::AlpakaBuiltInAtomic<alpaka::AtomicSub, unsigned int, alpaka::hierarchy::Grids, void>::atomic(unsigned int *, unsigned int)+0x480 in /home/lenz93/workspace/src/mallocMC-makeReservationPolicyOwning/alpaka/include/alpaka/atomic/AtomicUniformCudaHip.hpp:161
[...]
```
(sometimes also `out of bounds`).

Gonna investigate further in the near future but any comments and hints are welcome!

### LINKED ISSUE #252: Update to Catch2v3
Discussed offline: As part of the general overhaul planned, we should update to Catch2v3.

chillenzer:
This PR adds amongst a number of modernisations the new FlatterScatter creation policy.


examples/mallocMC_example01.cpp
```diff
@@ -26,90 +27,98 @@
...
 
-using ScatterAllocator = mallocMC::Allocator<
-    Acc,
-    mallocMC::CreationPolicies::Scatter<ScatterHeapConfig, ScatterHashConfig>,
-    mallocMC::DistributionPolicies::Noop,
-    mallocMC::OOMPolicies::ReturnNull,
-    mallocMC::ReservePoolPolicies::AlpakaBuf<Acc>,
-    mallocMC::AlignmentPolicies::Shrink<ShrinkConfig>>;
-
 ALPAKA_STATIC_ACC_MEM_GLOBAL int** arA;
```
psychocoderHPC, line 79: IMO this must be equal to https://github.com/alpaka-group/alpaka/blob/7e1b6ce24c5fadc9c58dd0d1dc59c022e96a087b/test/unit/mem/view/src/ViewStaticAccMem.cpp#L19

`alpaka::DevGlobal` is missing and the usage requires a `.get()` but it looks it runs and the `DevGlobal` is only required for SYCL. SO I am fine to keep it as it is.

src/include/mallocMC/creationPolicies/FlatterScatter.hpp
```diff
@@ -0,0 +1,456 @@
...
+         */
+        template<typename TAcc>
+        static void initHeap(auto& dev, auto& queue, auto* heap, void* pool, size_t memsize)
+        {
+            using Dim = typename alpaka::trait::DimType<TAcc>::type;
+            using Idx = typename alpaka::trait::IdxType<TAcc>::type;
+            using VecType = alpaka::Vec<Dim, Idx>;
+
+            auto poolView = alpaka::createView(dev, reinterpret_cast<char*>(pool), alpaka::Vec<Dim, Idx>(memsize));
+            alpaka::memset(queue, poolView, 0U);
```
psychocoderHPC: note: You mentioned last time offline that we use a kernel to initialize the heap but here the memset is still present. It is fine for this PR I would like to avoid that we use memset and a on device initialization.

src/include/mallocMC/creationPolicies/FlatterScatter.hpp
```diff
@@ -0,0 +1,456 @@
...
+            alpaka::wait(queue);
+
+            auto workDivSingleThread
+                = alpaka::WorkDivMembers<Dim, Idx>{VecType::ones(), VecType::ones(), VecType::ones()};
+            alpaka::exec<TAcc>(queue, workDivSingleThread, FlatterScatterAlloc::InitKernel{}, heap, pool, memsize);
```
psychocoderHPC, line 383: This sounds like the kernel which is performing the on device init equal to the memset above.

src/include/mallocMC/creationPolicies/FlatterScatter/AccessBlock.hpp
```diff
@@ -0,0 +1,823 @@
...
+        /**
+         * @brief Branch of getAvailableSlots for chunk sizes above the multi-page threshold. See there for details.
+         *
+         * @param chunkSize Would-be allocation size to test for.
+         * @return Number of allocations that would succeed with this size.
+         */
+        ALPAKA_FN_INLINE ALPAKA_FN_ACC auto getAvailableMultiPages(auto const& /*acc*/, uint32_t const chunkSize) const
+            -> uint32_t
+        {
+            // TODO(lenz): This is not thread-safe!
```
psychocoderHPC, line 344: we should not forget this todo and handle it in a future PR

src/include/mallocMC/creationPolicies/FlatterScatter/AccessBlock.hpp
```diff
@@ -0,0 +1,823 @@
...
+         * @param chunkSizeCache A memory location to store a local copy of the current chunk size. Used for
+         * optimisation by reducing the number of atomic lookups.
+         * @return true if the page is suitable and false otherwise
+         */
+        template<typename TAcc>
+        ALPAKA_FN_INLINE ALPAKA_FN_ACC auto thisPageIsSuitable(
+            TAcc const& acc,
+            uint32_t const index,
+            uint32_t const numBytes,
+            uint32_t& chunkSizeCache) -> bool
```
psychocoderHPC, line 645: only a note: this should be refactored in the future and return the value together with the current return value as tuple.

psychocoderHPC (changes requested):
offline discussed catch2 should be fetched on the need to not increase the repository size to much.

The rest looks good! ONly one small question for the heap init part is open.

src/include/mallocMC/creationPolicies/FlatterScatter.hpp
chillenzer: Oh, good catch! There is an `init` method for the heap but apparently it is unused. It's on my TODO list.

src/include/mallocMC/creationPolicies/FlatterScatter/AccessBlock.hpp
chillenzer: This is kind of a minor issue: The informational methods `getAvailableSlots` and its helpers provide information that cannot possibly be used in a thread-safe manner. So, calling them while other threads are write-accessing the same allocator is not a reasonable thing to do anyways. Mostly left it in for informational purposes but maybe I should drop the `TODO(lenz)` part. It is documented in the `Usage.md` document by the way.

examples/mallocMC_example01.cpp
chillenzer: Opened: https://github.com/alpaka-group/mallocMC/issues/272 to track this.

src/include/mallocMC/creationPolicies/FlatterScatter.hpp
chillenzer: Opened https://github.com/alpaka-group/mallocMC/issues/273 to track this.

src/include/mallocMC/creationPolicies/FlatterScatter.hpp
chillenzer: Oh, damn! It's much worse! The above memset is completely redundant. -.-

src/include/mallocMC/creationPolicies/FlatterScatter/AccessBlock.hpp
chillenzer: Opened https://github.com/alpaka-group/mallocMC/issues/274 to track this.

chillenzer:
Okay. I've surgically removed the Catch2 commit and replaced it with a git submodule. I've removed the redundant `memset` in the init method and will certainly do the initialisation benchmark again! Other than that I've opened issues to track your good suggestions.

psychocoderHPC (approved):

PR #275, opened 2025-01-15
Update cmake

chillenzer:
This is a major reorganisation of the repository including setting up proper CMake support. It is strongly inspired by [ModernCppStarter](https://github.com/TheLartians/ModernCppStarter) with some notable changes resulting from discussion with and preferences of @psychocoderHPC. As kind of a byproduct, also includes some modern features that are however not yet tweaked to production quality.

This PR is tested with [KitGenBench](https://github.com/ComputationalRadiationPhysics/KitGenBench) and [picongpu](https://github.com/ComputationalRadiationPhysics/picongpu). Hint for the review: The first commit removes alpaka and is quite unwieldy. Viewing the diff from the second commit on shows that it's really just cmake stuff and files moved.

I'll try to summarise the new layout and features:
- The main directory now provides a `CMakeLists.txt` that provides the library and optionally includes tests and examples via `mallocMC_BUILD_TESTING/EXAMPLES`.
- Directories containing source code are `include`, `examples` and `test`. The latter two are self-contained cmake subdirectories that can be built independently.
- All external source code is removed from the repository (except for a patched version of [CPM](https://github.com/TheLartians/CPM.cmake)).
- [CPM](https://github.com/TheLartians/CPM.cmake) is used for adding dependencies. This means that given internet access the library is fully self-contained and can download and build all dependencies itself if it must.
- Fine-grained control over the dependencies' origin is provided by variables of the form `mallocMC_USE_...` allowing `ON` (no downloading), `ON_ALLOW_FETCH` (default if required), `ON_ALWAYS_FETCH` (for best reproducibility), `AUTO` (default for optional dependencies) and `OFF`. Alternatively, the variable is allowed to contain a path pointing to the source location for this dependency.
- All details about the dependencies are specified in `cmake/package-lock.cmake`. Only ever touch this, if you want to change a version or the like. (Updating `alpaka` should be significantly less painful now.)
- Without internet connection one can simply download dependencies on another computer specifying the `-DCPM_SOURCE_CACHE=/some/convenient/path`, copy that to the offline system and point the same variable there for the build `-DCPM_SOURCE_CACHE=/where/you/copied/the/source/cache`. This is probably the closest we can get to offline support without shipping external code.
- In copying over from [ModernCppStarter](https://github.com/TheLartians/ModernCppStarter), we acquired a number of tools including platform-independent sanitizer, static analysis and code coverage support. At the time of writing, this is not yet properly configured and would need some tinkering with to produce useful output. (Currently, it can be run but most warnings I've seen are about alpaka and Catch2. It's hard to say at this point if there are useful messages about mallocMC among the noise.)


.github/workflows/ci.yml
```diff
@@ -1,5 +1,8 @@
 name: Continuous Integration
 on: [push, pull_request]
+env:
+  CPM_SOURCE_CACHE: ${{ github.workspace }}/cpm_modules
```
psychocoderHPC, line 4: out of interest: Is this variable also required in normal workflows or will the cache by default places somewhere relative to the build folder?

.github/workflows/ci.yml
```diff
@@ -18,26 +21,25 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/cache@v3
```
psychocoderHPC, line 24: why is all this configuration required?

CMakeLists.txt
```diff
@@ -1,89 +1,73 @@
...
+# header paths
+string(TOLOWER ${PROJECT_NAME}/version.hpp VERSION_HEADER_LOCATION)
+
+packageProject(
+  NAME ${PROJECT_NAME}
+  VERSION ${PROJECT_VERSION}
+  NAMESPACE ${PROJECT_NAME}
+  BINARY_DIR ${PROJECT_BINARY_DIR}
+  INCLUDE_DIR ${PROJECT_SOURCE_DIR}/include
+  INCLUDE_DESTINATION include/${PROJECT_NAME}-${PROJECT_VERSION}
```
psychocoderHPC, line 70: after installing, are all files in `include/malloc-3.0.0/mallocMC/*`?

cmake/tools.cmake
```diff
@@ -0,0 +1,74 @@
...
+# only activate tools for top level project
+if(NOT PROJECT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  return()
+endif()
+
+include(${CMAKE_CURRENT_LIST_DIR}/CPM_0.40.2.cmake)
+
+# enables sanitizers support using the the `USE_SANITIZER` flag available values are: Address,
+# Memory, MemoryWithOrigins, Undefined, Thread, Leak, 'Address;Undefined'
+if(USE_SANITIZER OR USE_STATIC_ANALYZER)
```
psychocoderHPC: the variables should have the prefix `mallocMC_`

cmake/tools.cmake
```diff
@@ -0,0 +1,74 @@
...
+    endif()
+
+    if(${CPPCHECK})
+      cppcheck(${CPPCHECK_ARGS})
+    endif()
+  endif()
+endif()
+
+# enables CCACHE support through the USE_CCACHE flag possible values are: YES, NO or equivalent
+if(USE_CCACHE)
```
psychocoderHPC: should be `mallocMC_USE_CCACHE`

examples/vectorAdd/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+cmake_minimum_required(VERSION 3.14...3.22)
+
+project(mallocMCExampleVectorAdd LANGUAGES CXX)
+
+# --- Import tools ----
+
+include(../../cmake/tools.cmake)
```
psychocoderHPC: should all `../..` start with `${CMAKE_CURRENT_LIST_DIR}/../..`?

test/multithreaded/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+project(mallocMCMultithreadedTests LANGUAGES CXX)
+
+# ---- Options ----
+
+option(ENABLE_TEST_COVERAGE "Enable test coverage" OFF)
+option(TEST_INSTALLED_VERSION "Test the version found by find_package" OFF)
+
+# --- Import tools ----
+
+include(../../cmake/tools.cmake)
```
psychocoderHPC: maybe `${CMAKE_CURRENT_LIST_DIR}/../..` ?

test/unit/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+cmake_minimum_required(VERSION 3.14...3.22)
+
+project(mallocMCUnitTests LANGUAGES CXX)
+
+# ---- Options ----
+
+option(ENABLE_TEST_COVERAGE "Enable test coverage" OFF)
```
psychocoderHPC: Those variables should start with `mallocMC_` to clearly show the users that mallocMC behaviour is changed.
This will also group all variables of mallocMC which helps finding all possible cmake switches.

psychocoderHPC (changes requested):
mostly variables without prefix `mallocMC_` were exposed.

.github/workflows/ci.yml
chillenzer: Offline discussion: This is not required but recommended. If not set, CPM will redownload for every new build directory. The downloaded files will be put into a local folder inside the build directory (if I recall correctly). The discussion led to another comment: We should document the option to not cache unstable dependencies (like the alpaka dev branch).

.github/workflows/ci.yml
chillenzer: Not required but supposed to slightly speed up CI runs by caching downloads. The CI here is only provisional at this point anyways. But concerning caching we should document the option to explicitly disable caching for unstable dependencies.

psychocoderHPC (approved):

PR #276, opened 2025-01-17
Fix submodules

chillenzer:
As discussed offline: The submodule was a bit loosely defined. It might be the cause of failures [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5246).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "thirdParty/catch2"]
-	path = thirdParty/catch2
-	url = https://github.com/catchorg/catch2
+  path = thirdParty/catch2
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "thirdParty/catch2"]
 	path = thirdParty/catch2
-	url = https://github.com/catchorg/catch2
+	url = https://github.com/catchorg/catch2.git
+  branch = v3.7.1
```
psychocoderHPC, line 4: wrong indention

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #282, opened 2025-02-05
Fix compilation error with multiple accelerators

chillenzer:
The inline assembly instructions for CUDA are `__device__`-only. `ALPAKA_FN_ACC` resolves to `__host__ __device__` unless we are in CUDA-only mode, so this leads to compilation failures. Fixing this by applying `__device__` by hand in those cases.

### DIFF
--- a/include/mallocMC/mallocMC_utils.hpp
+++ b/include/mallocMC/mallocMC_utils.hpp
@@ -131,7 +131,7 @@ namespace mallocMC
 
 #ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
     template<typename TDim, typename TIdx>
-    ALPAKA_FN_ACC inline auto smid(alpaka::AccGpuCudaRt<TDim, TIdx> const& /*acc*/) -> uint32_t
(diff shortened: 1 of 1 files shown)


include/mallocMC/mallocMC_utils.hpp
```diff
@@ -164,7 +164,7 @@ namespace mallocMC
 
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
     template<typename TDim, typename TIdx>
-    ALPAKA_FN_ACC inline auto lanemask_lt(alpaka::AccGpuHipRt<TDim, TIdx> const& /*acc*/)
+    ALPAKA_FN_ACC inline uint32_t lanemask_lt(alpaka::AccGpuHipRt<TDim, TIdx> const& /*acc*/)
```
psychocoderHPC: `ALPAKA_FN_ACC` should maybe changed to `__device__` too.

psychocoderHPC: **BUG**  `lanemask_lt()` return type must be `auto` because for AMD HIP the return type is 64bit and for CUDA it is a 32bit type.

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #287, opened 2025-02-06
Allocator takes Acc by tag

chillenzer:


include/mallocMC/allocator.hpp
```diff
@@ -113,6 +113,9 @@ namespace mallocMC
         using uint32 = std::uint32_t;
 
     public:
+        using Dim = alpaka::DimInt<1>;
+        using Idx = std::uint32_t;
```
psychocoderHPC, line 117: Note: This will limit the number of threads to 4 giga threads but this should be ok.

psychocoderHPC (approved):

PR #288, opened 2025-02-07
Add infrastructure and example for native CUDA

chillenzer:
Adds an example of using mallocMC in native CUDA applications and adds corresponding infrastructure. Along the way it fixes a cmake bug, namely that we forgot to link against alpaka. As there is a bug in alpaka's cmake forbidding this to succeed, we switch to a temporary branch until the main alpaka merges the fix.

After rebasing on top of the latest changes it turned out that our CI coverage is very poor and the reviewer apparently did not attempt manual testing on any of the previous PRs. So, the second commit fixes a plethora of inconsistencies that have built up lately.

Further minor feature requests/bug fixes crept in during offline discussion: All examples are now part of ctest, the tests enforce the serial accelerator they depend on and the native-cuda example is only activated if `alpaka_ACC_GPU_CUDA_ENABLE` is set.


chillenzer:
If https://github.com/alpaka-group/alpaka/pull/2476 gets merged first, we could update this PR to point to the alpaka dev again.

include/mallocMC/mallocMC.cuh
```diff
@@ -0,0 +1,184 @@
...
+
+    // Use the following in your native CUDA code and you are good to go! All alpaka-specific interfaces are patched
+    // away.
+    using detail::CudaAllocator;
+    using detail::CudaHostInfrastructure;
+    using detail::CudaMemoryManager;
+
+    namespace ReservePoolPolicies
+    {
+        // This is provided because the original ReservePoolPolicies::AlpakaBuf takes an alpaka::Acc as template
```
psychocoderHPC: Is this comment still true for the new mallocMC interface where we use a tag?

psychocoderHPC (changes requested):

include/mallocMC/mallocMC.cuh
chillenzer: Well, almost... it's still true that we provide this to make it alpaka-agnostic. It's no longer true that the alpaka-specific thing to patch out is the acc but only its tag. Anyways, I've updated the comment and added a sentence of clarification.

chillenzer:
As https://github.com/alpaka-group/alpaka/pull/2476 is already approved, I'm drafting this until it's merged, so that I can update to the latest dev of alpaka.

psychocoderHPC:
try to run examples and tests on our dev system fwk388 and I got a bunch of compile errors

```

spack load /gisjfoh
module load rocm-6.0.2
spack load cmake@3.29.4
export CXX=clang

cmake ../mallocMC -DmallocMC_BUILD_TESTING=ON -DmallocMC_BUILD_EXAMPLES=ON -Dalpaka_ACC_GPU_HIP_ENABLE=ON
```

[mallocMCCompileErrorsHIP.txt](https://github.com/user-attachments/files/18747187/mallocMCCompileErrorsHIP.txt)

psychocoderHPC:
`-Dalpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON` is required to run the tests. This should be set automatically by mallocMC

psychocoderHPC (approved):

PR #302, opened 2026-05-20
bug fix: adding data() function to return pointer to data of span

saeedelsayed:


saeedelsayed:
I had this error when trying to compile the native-cuda example:

`/mallocMC/examples/native-cuda/source/main.cu(91): error: class "mallocMC::span<uint64_t>" has no member "data"
      memoryManager.free(a.data());
                           ^
/home/saids/mallocMC/examples/native-cuda/source/main.cu(92): error: class "mallocMC::span<uint64_t>" has no member "data"
      memoryManager.free(b.data());
                           ^`

Please note that the `begin()` function is doing the same functionality as `data()` and can be used instead. However, `data()` is more expressive.

chillenzer:
> Please note that the `begin()` function is doing the same functionality as `data()` and can be used instead. However, `data()` is more expressive.

That's not quite true: They are identical on the implementation level but they have different semantics (by convention), compare [begin](https://en.cppreference.com/cpp/container/span/begin) and [data](https://en.cppreference.com/cpp/container/span/data). The former returns an iterator, the latter a pointer to the underlying data. As a C++ happens to fulfill the corresponding iterator concept, returning that pointer is semantically valid but not quite necessary for `begin` to have the correct semantics.

chillenzer (approved):
Well done! Thank you!

